PR #2, opened 2018-09-02
error : cannot resolve symbol cybersource

MUzayr:
import com.cybersource.ws.client.Client;          //error : cannot resolve symbol cybersource
import com.cybersource.ws.client.ClientException; //error : cannot resolve symbol cybersource
import com.cybersource.ws.client.FaultException;  //error : cannot resolve symbol cybersource

and many errors in the same file.


JacquesLeRoux:
Must be a problem between the chair and the desk

PR #3, opened 2019-03-18
Update GeoData_ID.xml

RoyIrwan:
Hi,
I believe geoTypeId for "ID-YO" should be "REGION" instead of "REGIOIN"


PierreSmits:
That is correct! Typos can easily happen.

Please be aware that, while the project is moving towards Git, we're still in the phase of discussion, documenting and planning.

Best regards,

JacquesLeRoux:
Fixed at 4ac9ae097e

PR #6, opened 2019-11-05
Identify target for set_multivalues() call from GLOBAL_LOOKUP_REF. 

danwatford:
Use jquery to find other fields to be updated based on element names.


JacquesLeRoux:
https://issues.apache.org/jira/browse/OFBIZ-11274 has been created for that, still pending, closing here

PR #13, opened 2020-02-11
Implemented: Add 'controlPath' attribute

PierreSmits:
(OFBIZ-11317)

added: Missed aspected in header.ftl in bluelight theme

### DIFF
--- a/themes/bluelight/template/Header.ftl
+++ b/themes/bluelight/template/Header.ftl
@@ -176,11 +176,11 @@ under the License.
                         </li>
                     </#if>
                 </#if>
-                <li class="user"><a href="/partymgr/control/viewprofile?partyId=${userLogin.partyId}${externalKeyParam!}">${userName}</a></li>
+                <li class="user"><a href="<@ofbizUrl controlPath="/partymgr/control">viewprofile?partyId=${userLogin.partyId}${externalKeyParam!}</@ofbizUrl>">${userName}</a></li>
               <#else>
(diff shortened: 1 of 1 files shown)


JacquesLeRoux:
Commited, thanks Pierre

PierreSmits:
What did you do?
After this I get in my local clone:

> git --no-optional-locks -c color.branch=false -c color.diff=false -c color.status=false -c diff.mnemonicprefix=false -c core.quotepath=false -c credential.helper=sourcetree fetch --prune OFBiz 
>fatal: Cannot fetch both refs/pull/13/head and refs/heads/pr/13 to refs/remotes/OFBiz/pr/13
>Completed with errors, see above

JacquesLeRoux:
As answered on dev ML
> I just made a commit and a push after cherry-picking your commit from PR13

JacquesLeRoux:
Maybe I should not have pushed to all branches but only trunk. I'll try that next time
I have checked this feature to push together R17 and R18 when backporting
13 h 55
So we need to delete the ‘refs/pull/13/head’ from remote, right?
Pierre Smits  [13 h 47]
That appears to be the way to go.
Jacques Le Roux  [13 h 50]
Done, how it goes on your side?
Both local and remote deleted
Pierre Smits  [13 h 51]
Let me check.
It seems ok now (and again). I see all available (and implemented PR) branches. As it should be.
13 h 56
OK, I guess it was the push ALL which created the issue (modifié)

PR #17, opened 2020-02-21
OFBIZ-10565 Removing unused UI labels

PierreSmits:
Cleaning up unused labels.
OFBIZ-9352 Accounting
OFBIZ-9367 HumanRes


JacquesLeRoux:
Hi Pierre,

I checked on trunk demo with the label manager for AccountingUiLabels.xml and then by hand locally. At least the 5 1st ones reported by the label manager are unused. Could you please check on your side and complete the patch?
TIA

PR #21, opened 2020-02-22
Fixed: Specified key was too long; max key length is 767 bytes for ProductPromoCodeEmail entity.(OFBIZ-5426)

pwverma:
Fixed: Specified key was too long; max key length is 767 bytes for ProductPromoCodeEmail entity.

(OFBIZ-5426)

The problem is in the entity model. An email address should not be used in the primary key - mainly because an email address is case-insensitive. A better design would be to use the email address contact mechanism ID in the primary key.

Done Following:
1. Changed Entity Name from ProductPromoCodeEmail to ProductPromoCodeContMech
2. Related Changes for the entity name change
3. Migration service to migrate old data

Thanks, Leon for the report and Adrian Crum, Jacques Le Roux, Ingo Wolfmayr, Deepak Dixit, Pierre Smits and Gil Portenseigne for the discussion and review.


applications/datamodel/entitydef/product-entitymodel.xml
```diff
@@ -3651,19 +3651,33 @@ under the License.
         <key-map field-name="productPromoCodeId"/>
       </relation>
     </entity>
+    <entity entity-name="ProductPromoCodeContMech"
```
dixitdeepak: Hi Pawan,
I think we can use ProdPromoCodeContactMech

applications/product/src/main/java/org/apache/ofbiz/product/migrate/MigrationServices.java
```diff
@@ -0,0 +1,73 @@
...
+import org.apache.ofbiz.service.ServiceUtil;
+
+public class MigrationServices {
+    public static final String module = MigrationServices.class.getName();
+
+    public static Map<String, Object> migrateProductPromoCodeEmail(DispatchContext dctx, Map<String, Object> context) {
+        Delegator delegator = dctx.getDelegator();
+        List<Object> errors = new LinkedList<>();
+        try {
+            List<GenericValue> productPromoCodeEmails = EntityQuery.use(delegator).from("OldProductPromoCodeEmail")
```
dixitdeepak: We can use queryIterator.

PR #22, opened 2020-02-22
OFBIZ-4035 change the id Attribute for input fields in the model macro form renderer from String to FlexibleStringExpander

danwatford:
Functionality to render a DOM element's id attribute using a FlexibleStringExpander expression already existed in Ofbiz via ModelFormField#getCurrentContainerId.

This PR includes changes in widget-form.xsd to document that the id-name attribute can use ${} notation. It also adds a small unit test to ensure that getCurrentContainerId is processing id-name as a FlexibleStringExpander expression.

### DIFF
--- a/framework/widget/dtd/widget-form.xsd
+++ b/framework/widget/dtd/widget-form.xsd
@@ -755,7 +755,18 @@ under the License.
                     <xs:documentation>Used to specify a javascript action that should be run based on an existing specified event.</xs:documentation>
                 </xs:annotation>
             </xs:attribute>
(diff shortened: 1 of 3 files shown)


mbrohl:
Hi Dan, it would be better to single out the documentation commit because this could be applied directly (as "Documented: ..."). The 2nd commit is already in PR 23 so could be removed (will need further review/decision).

mbrohl:
Please also use the commit message template (see issue comment).

danwatford:
Closing this PR as will create a new PR focused on the documentation commit per @mbrohl suggestion.

PR #23, opened 2020-02-22
OFBIZ-4035 Unit tests for MacroFormRenderer

danwatford:
This PR introduced changes to unit test MacroFormRenderer, but as this class has a lot of dependencies on static methods from various utility and environment classes, it seemed appropriate to make use of JMockit as a means to mock out static methods.

Ideally it wouldn't have been necessary to introduce a new mocking library, but given MacroFormRenderer's design it was a way to create some tests without having to build a very large number of brittle mockito mocks and lots of environmental setup code.


danwatford:
Closing as breaking submissions up into a number of more focused PRs.

PR #29, opened 2020-02-25
Fixed: Field Type definitions (OFBIZ-10955)

PierreSmits:
(OFBIZ-10955)

Adding missing field type definition for Apache Derby, MySQL, PostgreSQL


mbrohl (changes requested):
EDIT: will try to merge with amended commit message.

Hi Pierre,

can you please amend the commit message to Improved: ...
This is not a bugfix.

Thanks,
Michael

PR #32, opened 2020-02-25
OFBIZ-11418: Use FlexibleStringExpander in form widget lookup field field target parameters

danwatford:
Changed target-parameter attribute of the form widget's Lookup field to
use the FlexibleStringExpander type rather than String. This has been
done to allow reference to other form fields which may have had their
parameter-name attribute set to a FlexibleStringExpander expression.


JacquesLeRoux:
This has been committed with 08a371d0471a506e663c879f30299f424fc70921

PR #43, opened 2020-03-04
Implemented: OFBiz dockerization (OFBIZ-10407)

PierreSmits:
(OFBIZ10407) ]

Added: Dockerfile to establish an OFBiz container,
intended for evaluation and testing purposes.

To build the image:
docker build -t ofbiz/trunk:latest -f Dockerfile .

To run the container:
docker container run -ti -p 8009:8009 -p 8443:8443  --name OFBiz-Test ofbiz/trunk:latest


mbrohl (commented):
A few remarks:

- we should use the Adopt Open JDK as listed in the requirements
- the setup seems to add/install more than is needed
- what about the Apache Yetus comment, seems to belong to another Docker image?
- the used Ubuntu is outdated, we should use the 18.04 LTS release
- explicit Java installation can be removed by directly selecting the right Adopt Open JDK Docker image which makes also the ubuntu selection obsolete

PierreSmits:
@mbrohl,

Please put your comments in the associated ticket so that these can be seen by the entire community.

Mvg

Pierre

devops/Dockerfile
```diff
@@ -0,0 +1,84 @@
...
+ENV JAVA_HOME /usr/lib/jvm/java-8-openjdk-amd64
+
+
+####
+# Exposing default ports
+####
+EXPOSE 8009 8443
+
+WORKDIR /opt
+RUN git clone -q -o asf https://github.com/apache/ofbiz-framework.git ofbiz
```
ieugen, line 79: Having to do clone inside docker is not a good thing since you will have to provide credentials (for private repos).
Another way of handing this is to copy the current working directory inside docker and run the build from there:
Using multi stage docker images can help to make the images slimmer.

```
COPY .  /opt/ofbiz-source
WORKDIR /opt/ofbiz-source
RUN ./gradlew cleanAll loadAll
```

When building production images, once you compile ofbiz you can add a second FROM to copy the artifact and create the final image: https://docs.docker.com/develop/develop-images/multistage-build/ .

JacquesLeRoux:
This is abandonned, closing

PR #50, opened 2020-03-19
Fixed: remove CommonTheme dependency on Flatgrey (OFBIZ-11466)

PierreSmits:
Removing dependency of CommonTheme (theme.xml) on application.js in Flatgrey theme.
Code clean-up.


JacquesLeRoux:
Done locally by hand in trunk and R18. Not in R17, needs hands dirty, feel lazy this morning

PR #67, opened 2020-04-07
Improved: allowed hosts for Host Header Injection prevention (OFBIZ-11588)

PierreSmits:
(OFBIZ-11588)

added: local address 0.0.0.0
removed: non local FQDN names as they present a risk for the implementation


JacquesLeRoux:
As explained in the related Jira issue I close here because I see no clearly explained improvement and I don't see any Host Header Injection issues with the current value of host-headers-allowed.

Please reopen if you can better explain the reason of the argued improvement or/and demonstrate an Host Header Injection vulnerability with the current host-headers-allowed value, thanks.

PR #71, opened 2020-04-16
OFBIZ-11413 Improved Converted AccountServices from mini lang to groovy

wpaetzold:
(OFBIZ-11413)

Added ServiceDefs createPartyContactMechs and createPartyGroupRoleAndContactMechs.
Also converted createPartyGroupRoleAndContactMechs because it is used in AccountServies and needed more return values.


nmalin:
Loaded on trunk after some reword, thanks Wiebke

PR #74, opened 2020-04-20
OFBIZ-11595 Improved: Convert CostService.xml minilang to groovy

SebastianEcomify:


danwatford:
I've compared the minilang to groovy down to line 260 in CostServices.groovy and all looks good so far.

@SebastianEcomify were you able to test that you got same behaviour when using the services implemented in groovy compared to those implemented in minilang?

danwatford:
I've compared the rest of the minilang and groovy implementations and they look to be functionally the same.

danwatford (approved):

PR #76, opened 2020-04-25
Fixed: EntityQuery queryCount is throwing error with distinct method

pwverma:
(OFBIZ-11294)

Added missing support for selectFields in queryCount and all the subsequent methods.

### DIFF
--- a/framework/entity/src/main/java/org/apache/ofbiz/entity/GenericDelegator.java
+++ b/framework/entity/src/main/java/org/apache/ofbiz/entity/GenericDelegator.java
@@ -1620,13 +1620,17 @@ public class GenericDelegator implements Delegator {
         //TODO: add decrypt fields
         return eli;
     }
+    public long findCountByCondition(String entityName, EntityCondition whereEntityCondition,
+         EntityCondition havingEntityCondition, EntityFindOptions findOptions) throws GenericEntityException {
+        return findCountByCondition(entityName, whereEntityCondition, null, havingEntityCondition, findOptions);
+    }
 
     /* (non-Javadoc)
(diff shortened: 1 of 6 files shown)


framework/entity/src/main/java/org/apache/ofbiz/entity/GenericDelegator.java
```diff
@@ -1625,8 +1625,8 @@ public EntityListIterator findListIteratorByCondition(DynamicViewEntity dynamicV
      * @see org.apache.ofbiz.entity.Delegator#findCountByCondition(java.lang.String, org.apache.ofbiz.entity.condition.EntityCondition, org.apache.ofbiz.entity.condition.EntityCondition, org.apache.ofbiz.entity.util.EntityFindOptions)
      */
     @Override
-    public long findCountByCondition(String entityName, EntityCondition whereEntityCondition,
-            EntityCondition havingEntityCondition, EntityFindOptions findOptions) throws GenericEntityException {
+    public long findCountByCondition(String entityName, EntityCondition whereEntityCondition, Set<String> fieldsToSelect
```
dixitdeepak, line 1632: @verma-pawan I think we need to keep the existing method signature as well, and can pass null as fieldsToSelect to maintain backward compatibility.

pwverma: Thanks, @dixitdeepak totally agree. It slips off from my mind, on it.

pwverma: @dixitdeepak Done, please have look!

PR #77, opened 2020-04-25
Improved: Added support to consider non-default transaction timeout for Groovy and Simple event handlers.(OFBIZ-11607)

ravilodhi:


dixitdeepak (approved):

JacquesLeRoux (approved):
That's OK I'll commit

PR #91, opened 2020-04-30
Improved: Tax Authority Data for China (OFBIZ-11628)

PierreSmits:
(OFBIZ-11628)

added party and tax authority data for China


JacquesLeRoux:
That sound good to me, but I'd prefer to follow the way the data are organised. I mean to separate the blocks (party+PartyGroup+PartyRole) and (TaxAuthority+TaxAuthorityGlAccount) as it's currently done. It's easier when you need to "look" or "use" the data.

PierreSmits:
As it is in the patch and with various TaxAuthority parties in AccountingDemoData.xml. The inconsistency lies with several such parties and data from yore.

PierreSmits:
Correcting past data elements for demo purposes is (to be) done via OFBIZ-6976

PR #113, opened 2020-05-05
Improved: Convert EntitySyncServices.xml mini-lang to groovy (OFBIZ-11660)

PierreSmits:
(OFBIZ-11660)

removed: EntitySyncServices.xml
added: EntitySyncServices.groovy
modified: services.xml to reflect the change from minilang to groovy for:
- entitySyncPermissionCheck
- resetEntitySyncStatus
modified: webtools/controller.xml to reflect changed request-map and event type regarding reset of EntitySyncStatus
modified: webtools/widget/EntitySyncForms.xml to reflect changed target regarding reset of EntitySyncStatus


framework/entityext/servicedef/services.xml
```diff
@@ -447,7 +447,7 @@ under the License.
         <auto-attributes include="pk" mode="IN" optional="false"/>
     </service>
     <service name="entitySyncPermissionCheck" engine="simple"
-             location="component://entityext/minilang/EntitySyncServices.xml" invoke="entitySyncPermissionCheck">
+             location="component://entityext/groovyScripts/EntitySyncServices.groovy" invoke="entitySyncPermissionCheck">
```
surajkhurana, line 450: IMO, this can be made simpler using:

```
<service name="entitySyncPermissionCheck" engine="simple"
             location="component://common/minilang/permission/CommonPermissionServices.xml" invoke="genericBasePermissionCheck">
        <implements service="permissionInterface"/>
        <attribute name="primaryPermission" type="String" mode="IN" optional="true" default-value="ENTITY_SYNC"/>
        <attribute name="altPermission" type="String" mode="IN" optional="true"/>
    </service>
```

After this, you didn't need a groovy method.

PierreSmits: Thanks for the suggestion, @surajkhurana. I will look into that.

danwatford: Hi @surajkhurana, I was inclined to agree about your suggested change reducing the amount of code in ofbiz and was about to raise a jira issue to refactor other similar permission check services to use the Common Permission Services.

I then realised that doing so would couple the various permission check services to the implementation of the genericBasePermissionCheck, in this case a minilang implementation.

If we later want to re-implement genericBasePermissionCheck using a different engine (groovy, java, etc) then we will need to update all the application specific permission check services that depend on it.

In this case I think it is better to have application specific permission check services depend on the genericBasePermissionCheck service-interface and not the service-implementation.

framework/entityext/servicedef/services.xml
```diff
@@ -224,8 +224,8 @@ under the License.
             location="org.apache.ofbiz.entityext.synchronization.EntitySyncServices" invoke="cleanSyncRemoveInfo" auth="true" transaction-timeout="600">
         <description>Clean EntitySyncRemove Info - Generally should be run asynchronously after each sync run, or periodically run on a schedule</description>
     </service>
-    <service name="resetEntitySyncStatusToNotStarted" engine="simple"
-            location="component://entityext/minilang/EntitySyncServices.xml" invoke="resetEntitySyncStatusToNotStarted" auth="true" transaction-timeout="600">
+    <service name="resetEntitySyncStatus" engine="simple"
```
danwatford, line 227: engine needs to be set to "groovy"

framework/entityext/servicedef/services.xml
```diff
@@ -447,7 +447,7 @@ under the License.
         <auto-attributes include="pk" mode="IN" optional="false"/>
     </service>
     <service name="entitySyncPermissionCheck" engine="simple"
```
danwatford, line 449: Set engine to "groovy"

danwatford (changes requested):
Please set the engine on the two ported services to groovy.

danwatford:
Merged in PR #301

PR #123, opened 2020-05-09
Fixed: ListParty doesn't show middleName (OFBIZ-11681)

PierreSmits:
(OFBIZ-11681)

added: funtion to retrieve middleName from Person record and append it to the result


PierreSmits:
Implemented with PR #122

PR #128, opened 2020-05-10
Improved: accounting component - have configuration options in db (OFBIZ-6189)

PierreSmits:
(OFBIZ-6189)

added: SystemProperty records in AccountingSystemPropertyData.xml
added: reference of AccountingSystemPropertyData.xml to load at seed loading moment


PierreSmits:
Implemented via the merge of PR388

PR #130, opened 2020-05-10
Improved: Converted CustomTimePeriod related services from simple to entity-auto

surajkhurana:
(OFBIZ-11689)
Thanks Pierre for reporting.


nmalin:
Loaded on trunk thanks

PR #138, opened 2020-05-14
Fixed: Themes provides link to partymgr in headers (OFBIZ-11703)

PierreSmits:
(OFBIZ-11703)

removed: <a href...> to viewprofile


JacquesLeRoux:
This can be useful for allowed users, others can't that's all. A message would help, but that's not what did Pierre. He simply removed the possible fonctionnality, closing

PR #139, opened 2020-05-15
Improved: gradle initialisation scripts (OFBIZ-11707)

PierreSmits:
renamed: scripts to init-gradle
moved: location of init-gradle script for linux-type OSes
improved: INSTALL doc regarding renamed scripts


JacquesLeRoux:
We don't need that

PR #144, opened 2020-05-17
Improved: removed AR and AP Webapps from Accounting (OFBIZ-10552)

PierreSmits:
(OFBIZ-10552)


PierreSmits:
Are we still going to go through with the removal of the AR and AP from the accounting webapps?

PR #158, opened 2020-05-23
[Improved] Convert createPaymentApplication service from mini-lang to groovy DSL (OFBIZ-11481)

priyasharma1:
(OFBIZ-11481)

Done the following:
- removed mini-lang version of the service
- added groovy version
- updated the service definition


applications/accounting/groovyScripts/payment/PaymentServices.groovy
```diff
@@ -134,3 +136,77 @@ def updatePaymentContent() {
...
+        if (invoice.billingAccountId) {
+            paymentAppl.billingAccountId = invoice.billingAccountId
+        }
+    }
+
+    if (parameters.toPaymentId) {
+        // get the to payment and check the parent types are compatible
+        GenericValue toPayment = from("Payment").where("paymentId", parameters.toPaymentId).queryOne()
+        if (toPayment) {
+            toPaymentType = from("PaymentType").where("paymentTypeId", toPayment.paymentTypeId).queryOne()
```
dixitdeepak: we can use getRelated method with cache.

applications/accounting/groovyScripts/payment/PaymentServices.groovy
```diff
@@ -134,3 +136,77 @@ def updatePaymentContent() {
...
+        }
+        paymentType = from("PaymentType").where("paymentTypeId", payment.paymentTypeId).queryOne()
```
dixitdeepak: Use cache.

applications/accounting/groovyScripts/payment/PaymentServices.groovy
priyasharma1: Thanks, Deepak.
As I was trying to make this improvement, I noticed that there two variables are never used i.e. paymentType and toPaymentType.
Will it be fine to skip these two entity operations if they are not used?

pwverma: Hi Priya,

Yes, If a variable is never used we can remove it safely.

priyasharma1: Thanks, Pawan for the verification. I have removed both those lines.
Though I have tested it please let me know if this causes any issue.

nmalin:
Thanks @priyasharma1 , I merged on trunk from my local after groovy simplification.

PR #166, opened 2020-05-29
Fixed: Create Payment service throws an exception when receiving payment on order (OFBIZ-11754)

priyasharma1:
(OFBIZ-11754)

- Since the MODULE variable could not be found, the service was throwing error, thus removed the variable
- Instead used the groovy "logInfo" method directly to print the log, which has access to the module variable implicitly.
- Removed the unused import, as Debug class is no longer needed here.
- Added these changes on the ContentServices.groovy as well, as it was implemented in the same pattern.

Thanks: Pawan Verma & Suraj Khurana for the guidance and discussion

### DIFF
--- a/applications/content/groovyScripts/content/ContentServices.groovy
+++ b/applications/content/groovyScripts/content/ContentServices.groovy
@@ -28,10 +28,8 @@ import org.apache.ofbiz.service.GenericServiceException;
 
 import org.apache.ofbiz.service.ModelService
 import org.apache.ofbiz.service.ServiceUtil
-import org.apache.ofbiz.base.util.Debug
 import org.apache.ofbiz.base.util.UtilDateTime
 
-MODULE = "ContentServices.groovy"
 def createTextAndUploadedContent(){
     Map result = success()
 
@@ -115,7 +113,7 @@ def createContentAlternativeUrl() {
                             dataResourceId = serviceResult.dataResourceId
                         }
                     } catch (GenericServiceException e) {
-                        Debug.logInfo(e, MODULE)
+                        logInfo(e)
                     }
                     if (dataResourceId) {
                         serviceContext.clear()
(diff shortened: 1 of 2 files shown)


applications/content/groovyScripts/content/ContentServices.groovy
```diff
@@ -115,7 +113,7 @@ def createContentAlternativeUrl() {
                             dataResourceId = serviceResult.dataResourceId
                         }
                     } catch (GenericServiceException e) {
-                        Debug.logInfo(e, MODULE)
```
pwverma, line 118: Hi @priyasharma1 

As we are logging exception here, Shouldn't it be converted to logError()? WDYT?

JacquesLeRoux:
Also why not using logError (from GroovyBaseScript.groovy) instead of  Debug.logError?

I even wonder if we should change that everywhere. But then a Jira and maybe an agreement on dev ML would be required (later does not seem required to me)

applications/content/groovyScripts/content/ContentServices.groovy
priyasharma1: Thanks @verma-pawan 
I have made the changes as per your suggestion.

Earlier, since the original XML service used break-on-error="false" without any logs at all, so I could not decide whether to log error or info.

So thank you for your inputs in clearing up my doubt.

priyasharma1:
> Also why not using logError (from GroovyBaseScript.groovy) instead of Debug.logError?
> 
> I even wonder if we should change that everywhere. But then a Jira and maybe an agreement on dev ML would be required (later does not seem required to me)

Yes, Jacques, I have used the same in the updated commit.
Thanks!

JacquesLeRoux:
Ha Pawan already created ttps://issues.apache.org/jira/browse/OFBIZ-11762 (" Use GroovyBaseScript's logging utility methods instead of using Debug in each Groovy files") :)

pwverma (approved):
Changes looks good to me, +1 for merge (y)

PR #184, opened 2020-06-02
Fixed: Overview of shipments shows wrong link to shipment (OFBIZ-11787)

PierreSmits:
See commit message.

### DIFF
--- a/applications/product/widget/facility/ShipmentForms.xml
+++ b/applications/product/widget/facility/ShipmentForms.xml
@@ -70,12 +70,12 @@ under the License.
                 <field-map field-name="viewSize" from-field="viewSize"/>
             </service>
         </actions>
-        <field name="shipmentId" title="${uiLabelMap.ProductShipmentId}" widget-style="buttontext">
-            <hyperlink description="${shipmentId}" target="EditShipment">
+        <field name="shipmentId" title="${uiLabelMap.CommonShipment}" widget-style="buttontext">
+            <hyperlink description="${shipmentId}" target="ViewShipment">
                 <parameter param-name="shipmentId" from-field="shipmentId"/>
             </hyperlink>
         </field>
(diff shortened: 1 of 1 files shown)


applications/product/widget/facility/ShipmentForms.xml
```diff
@@ -70,12 +70,12 @@ under the License.
                 <field-map field-name="viewSize" from-field="viewSize"/>
             </service>
         </actions>
-        <field name="shipmentId" title="${uiLabelMap.ProductShipmentId}" widget-style="buttontext">
-            <hyperlink description="${shipmentId}" target="EditShipment">
+        <field name="shipmentId" title="${uiLabelMap.CommonShipment}" widget-style="buttontext">
```
pwverma, line 73: Hi @PierreSmits 
I think we can use CommonShipmentID here, WDYT?

PierreSmits:
Context is everything. And a Shipment is represented with its Id. So why not keep it simple and lose the Id in the label when the obvious is obvious?

PierreSmits:
I am confident, that many don't need the inclusion of Id in the label when the context is clear and unambiguous. Similarly like not including Id on status labels (uiLabelMap.*StatusId).

pwverma:
> Context is everything. And a Shipment is represented with its Id. So why not keep it simple and lose the Id in the label when the obvious is obvious?

Yes, I agree, Context is everything. But we need to maintain consistency throughout the system. We can change this pattern all over the application in a separate effort.

PierreSmits:
Indeed. And I have suggested such changes regarding simplification of labels (and application thereof) a few times in the past, but that was met with too much obstruction from some contributors. Mainly being about patches being either to small or to big. And that is a pity.

Consistency (and ease of use) should be key, but is nevertheless a big endeavour that will take effort to implement. See my attempts to get to a consistent way of having a Main Action Menu (for the creation process of the most important objects) in each web app. (see other PRs).

pwverma:
I've merged changes manually so closing this PR.

PR #185, opened 2020-06-03
WIP: [Improved]: Use GroovyBaseScript's logging utility methods instead of using Debug (OFBIZ-11762)

priyasharma1:
(OFBIZ-11762)

- Replaced the debug.log methods with the respective groovy utility method
- Enhanced the GroovyBaseScript to entertain fetching of the current class name as the module name for logging
- Overloaded the logError method to entertain throwable object for better logging
- Removed unused import statements and module variables.

Thanks: Pawan & Jacques

### DIFF
--- a/applications/accounting/groovyScripts/invoice/InvoiceServices.groovy
+++ b/applications/accounting/groovyScripts/invoice/InvoiceServices.groovy
@@ -30,7 +30,7 @@ def getNextInvoiceId() {
 
     // try to find PartyAcctgPreference for parameters.partyId, see if we need any special invoice number sequencing
     GenericValue partyAcctgPreference = from('PartyAcctgPreference').where(parameters).queryOne()
-    if (Debug.infoOn()) Debug.logInfo("In getNextInvoiceId partyId is [${parameters.partyId}], partyAcctgPreference: ${partyAcctgPreference}", 'InvoiceServices.groovy')
+    if (Debug.infoOn()) logInfo("In getNextInvoiceId partyId is [${parameters.partyId}], partyAcctgPreference: ${partyAcctgPreference}")
 
     String customMethodName = null;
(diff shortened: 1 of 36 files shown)


JacquesLeRoux:
Thanks Priya,

It's OK with me but for EditSurveyQuestions.groovy  OpenOrderItemsReport.groovy ComputeProductSellThroughData.groovy CountFacilityInventoryByProduct.groovy  ProductPromoCondServices.groovy  FetchLogs.groovy  in which, in a least one case, module is used and should not.

This said I find annoying that Github doest not allow to increase the width of the view when reviewing. Since we agreed on 150 chars lines width it's really an issue. I could not find an easy way to increase the width of the view when reviewing.

mbrohl:
I am not able to check right now but are we sure that this change does not introduce unwanted changes? Does it respect the log configurations with log4j?

priyasharma1:
Thanks, Jacques

Actually, in these files, I found the message to be like Debug.logError(e,  "Failure in " + module)
Thus I did not omit the "module" in the error message.
IMO module will still point to the current class name, and so it won't cause any loss.

Or, we can alter the error message, if you agree to that.
Please let me know your views.

priyasharma1:
> I am not able to check right now but are we sure that this change does not introduce unwanted changes? Does it respect the log configurations with log4j?

Thanks for the review, Michael. :)
Please feel free to test it thoroughly and let me know in case you have any suggestions or improvements.

We have only made changes to use the groovy utility methods, which internally invokes the Debug logging methods. thus there should be no conflicts with the log configurations.

JacquesLeRoux:
Hi Priya,

Thanks, got it, it's OK with me. I'll test and merge if all is OK

JacquesLeRoux:
Hi Michael,

I tested and there is no problem with your concerns, merging

PR #188, opened 2020-06-04
Improved: flatgrey - simplified uris (OFBIZ-11795)

PierreSmits:
adjusted the main menu ftl to work with simplified uris


JacquesLeRoux:
Closing here, we can't push related PRs one by one. We need the whole thing for OFBiz to correctly work with the related changes. In other words, it's all or nothing. It seems possible....

PR #189, opened 2020-06-05
Improved: rainbow - simplified uris (OFBIZ-11797)

PierreSmits:
implement changes to the rainbow theme to provide simplified uris in the main menu.


JacquesLeRoux:
Closing here, we can't push related PRs one by one. We need the whole thing for OFBiz to correctly work with the related changes. In other words, it's all or nothing. It seems possible....

PR #192, opened 2020-06-06
Improved: accounting - simplified uri (OFBIZ-11800)

PierreSmits:


JacquesLeRoux:
Closing here, I'll handle it by hand, there is no reason to remove SameSiteFilter, that would be a rough error

PR #193, opened 2020-06-06
Improved: manufacturing - simplified uri (OFBIZ-11801)

PierreSmits:


JacquesLeRoux:
After few hours trying, won't do

PR #197, opened 2020-06-10
Improved: humanres - simplified uri (OFBIZ-11814)

PierreSmits:


JacquesLeRoux:
Closing here, I'll handle it by hand, there is no reason to remove SameSiteFilter, that would be a rough error, see OFBIZ-11470

PR #199, opened 2020-06-10
Improved: sfa - simplified uri (OFBIZ-11816)

PierreSmits:


JacquesLeRoux:
After few hours trying, won't do

PR #203, opened 2020-06-14
[Implemented] Introduce the option for specifying Check (Reference) Number on order entry (OFBIZ-7377)

priyasharma1:
Done the following:

- Added a field on the UI to allow the input of the check reference number, focus on the filed will also check the related payment method.
- Added logic to set this reference number on the cartpaymentinfo map for further processing
- Added logic to fetch the reference number from the order preference record and apply it on the payment.

(OFBIZ-7377)


pwverma (approved):
Changes look good to me, I've tested it is working well. Thanks, Priya!

PR #204, opened 2020-06-18
added a report uri and unsafe-inline to the csp header.

alexbodn:
i understand the policy is against unsafe-inline,
but as a matter of fact there are as lot of inline js/css,
and part of the js is adding additional remote js.
the report uri is implemented in a plugin with this purpose.

Improved:
Implemented:
Documented:
Completed:
Reverted:
Fixed:
(OFBIZ-)

Explanation

Thanks:


JacquesLeRoux:
Hi Alex,

I agree there are still inline js/css, even if most are handled correctly. We have related Jiras to eventually use require.js: https://issues.apache.org/jira/browse/OFBIZ-9465 https://issues.apache.org/jira/browse/OFBIZ-9976. Even if they not yet got much attention we still hope to work on it.

So you can do that in your own project, we won't in official repo. If you disagree you may start a thread on the [dev ML](dev@ofbiz.apache.org). Else please close.

mbrohl:
@alexbodn Can you please close this PR?

JacquesLeRoux:
Hi Michael,

It seems Alex is no longer interested in OFBiz, closing...

PR #214, opened 2020-07-13
Implemented: New service definition attribute "verb" to denote the corresponding HTTP method (OFBIZ-11328)

girishvasmatkar:
Implemented: New service definition attribute "action" to denote the corresponding HTTP method (OFBIZ-11328)

Explanation
In order for exportable services in OFBiz to be called via REST interface, it is necessary for the service to tell what HTTP verb should be used to call it. This implementation adds the attribute and makes it available in ModelService class via property **action**.

Thanks:
Girish


girishvasmatkar:
I'm keeping it open for review by the reviewers. I will merge it during the upcoming weekend.

PierreSmits:
If we're talking about an action here, why don't we call it 'action'?

girishvasmatkar:
@PierreSmits : "action" makes more sense. I'll change the attribute name and then merge. Thanks.

PR #215, opened 2020-07-19
Improved : Convert createRoleType service to entity-auto from Java and update its service calls (OFBIZ-11894)

priyasharma1:
(OFBIZ-11894)

- Removed the java implementation
- Converted the service to entity-auto
- Updated the service call in Party tests where this service was used.


pwverma (approved):
LGTM

PR #216, opened 2020-07-20
Improved: Convert createPartyRelationshipType service to entity-auto from Java and update its service calls (OFBIZ-11895)

priyasharma1:
(OFBIZ-11895)

- Updated the service definition to convert it into entity-auto
- Added permission check service as used in the Java implementation
- Removed the java implementation of the services


pwverma (approved):
Looks good to me

PR #221, opened 2020-07-30
Improved: WidgetWorker should not write generated html to Appendable (OFBIZ-11907)

danwatford:
Refactoring of WidgetWorker so that it generates URI and JSoup Element
objects to represent created URLs, hidden forms and anchor tags. This
replaces the previous approach where WidgetWorker would write string
representations of the URLS, hidden forms and anchor tags directly to an
Appendable object passed to it.

Callers to WidgetWorker have been modified to render the new objects
created by WidgetWorker to their relevant I/O.


danwatford:
@JacquesLeRoux New PR created after applying the code changes to trunk HEAD. Patching should be cleaner this time.

JacquesLeRoux:
Thanks Daniel,

The patch applies, I'll review ASAP...

PR #223, opened 2020-08-06
Improved: Make all related Cancel/Done buttons return to backHome ie. party profile page (OFBIZ-11943)

priyasharma1:
Explanation

- Updated the targets to 'backHome' so as to return back to the profile page
- Added roleTypeId on other instances as it is because of this missing parameter that the SFA profile page is not visible.

(OFBIZ-11943)


JacquesLeRoux:
Closing as it was committed as a slightly modified patch

JacquesLeRoux:
We need another Jira for similar issues on the same page

PR #226, opened 2020-08-19
Improved: Form-Fields of certain types won't render IDs (OFBIZ-11957)

dbalkir:
Improved: Form-Fields of certain types won't render IDs (OFBIZ-11957)

Added the ability to add IDs to OFBiz-Form-Fields like Hyperlink and
Submit to access them via JavaScript or similar things


dbalkir:
Somehow this was missing an important part

PR #232, opened 2020-08-23
OFBIZ-11900: Improved: MacroFormRenderer refactoring of label, display and text fields

danwatford:
New RenderableFtl elements to represent pre-rendered FTL strings and FTL
macro calls. RenderableFtl elements are able to render themselves to
strings which are processed as an FTL template by the FtlWriter class.

For labels, display fields and text fields, MacroFormRenderer no longer
generates FTL to write to a template itself, but instead calls
RenderableFtlFormElementsBuilder to create corresponding RenderableFtl
elements which are then processed by FtlWriter. 

This is a WIP to reduce complexity in MacroFormRenderer. More changes will follow to move generation of FTL out of MacroFormRenderer to RenderableFtlFormElementsBuilder and other classes.


danwatford:
This PR included use of Lombok which has not been accepted into the Ofbiz project.
Closing in favour or #237 which makes the same improvements but with Lombok removed.

PR #236, opened 2020-10-26
Improved: Remove _PREVIOUS_REQUEST_ Session Attribute on non-authentication pages (OFBIZ-12047)

ingo-richter:
Improved: Remove _PREVIOUS_REQUEST_ Session Attribute on non-authentication pages (OFBIZ-12047)

Added removal of the _PREVIOUS_REQUEST_ attribute when requesting non-authenticated sites and moved targetRequestUri handling to a accommodate this change

### DIFF
--- a/framework/webapp/src/main/java/org/apache/ofbiz/webapp/control/RequestHandler.java
+++ b/framework/webapp/src/main/java/org/apache/ofbiz/webapp/control/RequestHandler.java
@@ -376,13 +376,6 @@ public class RequestHandler {
 
         // Grab data from request object to process
         String defaultRequestUri = RequestHandler.getRequestUri(request.getPathInfo());
-        if (request.getAttribute("targetRequestUri") == null) {
-            if (request.getSession().getAttribute("_PREVIOUS_REQUEST_") != null) {
-                request.setAttribute("targetRequestUri", request.getSession().getAttribute("_PREVIOUS_REQUEST_"));
-            } else {
-                request.setAttribute("targetRequestUri", "/" + defaultRequestUri);
-            }
-        }
 
         String requestMissingErrorMessage = "Unknown request ["
                 + defaultRequestUri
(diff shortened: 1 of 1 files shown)


mbrohl:
Looks good to me and is known to work in custom projects.
Any objections to get this into the codebase?

mbrohl (approved):

JacquesLeRoux:
Hi Michael,

It would be good to have an explanation of the reason of this change

ingo-richter:
Hi Jacques,

I tried to explain the reasons in the corresponding JIRA issue but found the problem I mentioned (navigation problems while navigating the site without a login and then logging in) does not exist on the demo front end due to the implementation of the login.

It is possible to implement a login button using the current methods that automatically sends the user to the previously viewed page (which we have done in custom projects). Without the commit provided the last visited page will not be cleared on appropriate navigation, resulting in a mismatch between where the user currently is on the site and the site saved as the last visited. This results in said login leading to the (falsely identified) last visited site (which required authorization) instead of the current site (if it does not require authorization).

For example:
1. Navigate to a profile page requiring authorization.
2. Navigate to an FAQ site which does not require authorization.
3. The login button should now lead, upon authorization) back to the FAQ site. Instead the browser is send to the profile page since it was the last saved request requiring authorization.

I hope this explanation clears up the issue.

Best regards,
Ingo

JacquesLeRoux:
Thanks Ingo,

I should have a 2nd look at OFBIZ-12047 (I did when you created it) before asking, but this explanation is more convincing :)

JacquesLeRoux (approved):

mbrohl:
@ingo-koenemann please provide a new PR with modified commit message "Fixed: ..." because we have classified this as a bug now. Thank you!

ingo-richter:
Ich created a new PR #262.

mbrohl:
Closed for #262

PR #244, opened 2020-11-28
Fixed: Single product tile is displayed on one row in product search (OFBIZ-12045)

isdixit:
Cause: Wrong use of bootstrap grid classes.

Fixed: Used row class as a parent to col-* class while rendering the products.


isdixit:
Changes have been moved to new PR [!245.](https://github.com/apache/ofbiz-framework/pull/245)

PR #245, opened 2020-11-28
Fixed: Single product tile is displayed on one row in product search (OFBIZ-12045).

isdixit:
Cause: Wrong use of bootstrap grid classes.
Fixed: Used row class as a parent to col-* class while rendering the products.


danwatford (approved):

JacquesLeRoux:
Thanks: Arpit Mor for report, Saurabh Dixit for patch, Daniel Watford for test

PR #251, opened 2021-01-14
Improved: Add extra classpath directories for ofbiz.tar/zip

ieugen:
(OFBIZ-12136)

Allows people to add configuration files and jars (database drivers) when
using the binary distribution.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1106,3 +1106,15 @@ new File("${gradle.getGradleUserHomeDir().getAbsolutePath()}/daemon/${gradle.get
         it.delete()
     }
 }
+
+tasks.startScripts {
+    doLast {
+        // Alter the start script for Unix systems.
+        unixScript.text =
+                unixScript.text.replace('CLASSPATH=$APP_HOME/lib','CLASSPATH=$APP_HOME/config/:$APP_HOME/lib-extra/*:$APP_HOME/lib')
+        // Alter the start script for Windows systems.
+        windowsScript.text =
+                windowsScript.text.replace('CLASSPATH=%APP_HOME%\\lib',
+ 'CLASSPATH=%APP_HOME%\\conf\\;%APP_HOME%\\lib-extra\\*;%APP_HOME%\\lib')
+    }
+}


danwatford:
Would the following work? It would mean we wouldn't need to manipulate the text content of the two different types of start script:

```
tasks.startScripts {
    classpath = files('lib/conf/', 'lib-extra/*') + classpath
}
```

ieugen:
> Would the following work? It would mean we wouldn't need to manipulate the text content of the two different types of start script:
> 
> ```
> tasks.startScripts {
>     classpath = files('lib/conf/', 'lib-extra/*') + classpath
> }
> ```

I did not try but I think Gradle will resolve those during build (the files function). 
We need the scripts to take the classpath when the script is run and not try to resolve the `lib-extra/*` path.

danwatford:
@ieugen , yes you're right.
Setting the classpath property for startScripts did indeed try to resolve all the wild card (*) entries at build time and place them under /lib in the distribution. Never mind.

danwatford:
@ieugen  I haven't been able to test these changes successfully on Windows with git-bash.

I created a distZip using the PR.
Unzipped to c:\dev\tmp\zip
In git bash:
```
cd /c/dev/tmp/zip
bin/ofbiz
```

The above launched ofbiz and I was able to access https://localhost:8443/partymgr/control/login successfully.

I then copied start.properties and url.properties to /c/dev/tmp/zip/ofbiz/config.
Edited start.properties to have `ofbiz.log.dir=runtime/logs2`
Edited url.properties to have `port.https=9443`
Created directory /c/dev/tmp/zip/ofbiz/runtime/logs2

I then ran ofbiz using `bin/ofbiz` and attempted to access `https://localhost:9443/partymgr/control/login` whereas I could still connect to port 8443.

Also no files were written to runtime/logs2.

Can you give an example of a property you were able to override by using the new config directory?

ieugen:
I don't have Windows to test that on.
I just made similar changes to CLASSPATH.
I read ofbiz.bat, saw how the CLASSPATH was used and added the same options.

I did test on linux and you can test too using docker:
I use these commands to run OFBiz on my Raspnerry PI with postgresql db running in another container. 

```
    # Data loading
    docker run --rm \
        --name ofbiz-netdava \
        --net=host \
        --volume /opt/ofbiz/entityengine.postgres.xml:/opt/ofbiz/config/entityengine.xml:ro \
        --volume /opt/ofbiz/security.properties:/opt/ofbiz/config/security.properties:ro \
        --volume ofbiz-netdava:/opt/ofbiz/runtime \
        ieugen/ofbiz:trunk-8 --load-data

    docker run --detach \
        --name ofbiz-netdava \
        --restart unless-stopped \
        --net=host \
        --volume /opt/ofbiz/entityengine.postgres.xml:/opt/ofbiz/config/entityengine.xml:ro \
        --volume /opt/ofbiz/security.properties:/opt/ofbiz/config/security.properties:ro \
        --volume ofbiz-netdava:/opt/ofbiz/runtime \
        ieugen/ofbiz:trunk-8
```

danwatford:
Hi @ieugen , I've been able to successfully test overriding properties by placing files in the dist's config directory.

I tested in an openjdk-8 container using:
`docker run -it --rm --publish 8080:8080 --publish 8443:8443 --mount type=bind,source=$PWD/build/distributions,destination=/distributions openjdk:8 bash -c "tar xf /distributions/ofbiz.tar; cd /ofbiz; bash"`

I placed a copy of security.properties file in /ofbiz/config and set `password.length.min=10`

Data was loaded with `bin/ofbiz -l` and then ofbiz started with `bin/ofbiz`

I attempted to change the admin user's password to a single character value and was correctly shown an error message stating minimum password length was 10 characters, proving that the config override had worked.

I then repeated the test using security.properties in windows which also passed. It seems url.properties and start.properties were not good candidates for overriding, perhaps they are loaded in an unusual way.

danwatford:
@ieugen I 'squashed and merged' using GitHub for this PR and the original author (you) still gets recorded against the commit.

ieugen:
@danwatford : start.properties is in package org/apache/ofbiz/base/start/start.properties  .
So you will need a full directory path for it. 

I don't know in which package url.properties should be.

PR #252, opened 2021-01-19
Fixed: Use data url scheme for logo in CompanyHeader (OFBIZ-7327)

danwatford:
(OFBIZ-7327)

Use a data URL rather than a regular URL for company logos when a
logo has been specified as a Party Content item. This works around the
problem of Apache FOP failing to retrieve content from HTTPS URLs.


applications/order/groovyScripts/order/CompanyHeader.groovy
```diff
@@ -135,12 +136,22 @@ if (!partyId) {
 }
 
 // the logo
-partyGroup = from("PartyGroup").where("partyId", partyId).queryOne()
+GenericValue partyGroup = from("PartyGroup").where("partyId", partyId).queryOne()
 if (partyGroup) {
-    partyContentWrapper = new PartyContentWrapper(dispatcher, partyGroup, locale, EntityUtilProperties.getPropertyValue("content", "defaultMimeType", "text/html; charset=utf-8", delegator))
-    partyContent = partyContentWrapper.getFirstPartyContentByType(partyGroup.partyId , partyGroup, "LGOIMGURL", delegator)
+    GenericValue partyContent = PartyContentWrapper.getFirstPartyContentByType(partyId, partyGroup, "LGOIMGURL", delegator)
```
dixitdeepak, line 141: I think here we can use getPartyContentAsText directly, it will give you the url stored in dataResource, also it will take care to read data from other supported types as well.

danwatford: Using getPartyContentAsText didn't work for me. My various attempts at using PartyContentWrapper, ContentWork and DataResourceWorker to render a URL for me didn't work out. I ended up just getting the Content ID returned as a String.

dixitdeepak: Could you please share the dataResouce data?
It should work properly

danwatford: I can take another look, but before I do so what sort of URL would you expect from getPartyContentAsText?

If the URL is of the form https://localhost/content/control/stream?contentId=10001 then this won't solve the problem. I still need to read the binary data of the content so that I can turn it into a base64 encoded data-uri.

dixitdeepak: This will give the file path if data setup properly 
It depends upon combination of following fields 
- Data Template Type ID 
- Data Resource Type ID

danwatford: Hi @dixitdeepak , I still just get a string version of the content ID.

After adding a new logo and assigning to the Company party as the logo image I end up with the following GenericEntities:

(I appreciate the following is difficult to read so please let me know of any preferred way to get and present this sort of information.)

dataResource.asImmutable().toString():
```
{dataResourceName=TestLogo.png, lastUpdatedStamp=2021-01-19 13:57:11.498, surveyId=null,
dataTemplateTypeId=NONE, lastModifiedDate=2021-01-19 13:57:10.159, createdTxStamp=2021-01-19 13:57:09.983,
createdStamp=2021-01-19 13:57:11.176, mimeTypeId=image/png, lastUpdatedTxStamp=2021-01-19 13:57:09.983, 
dataCategoryId=PERSONAL, surveyResponseId=null, characterSetId=null, dataSourceId=null,
lastModifiedByUserLogin=admin, createdDate=2021-01-19 13:57:10.159, statusId=CTNT_PUBLISHED,
localeString=null, relatedDetailId=null, dataResourceId=10030, isPublic=N, 
objectInfo=C:/dev/floss/ofbiz/ofbiz-framework/runtime/uploads/1610721557185/10030.png,
dataResourceTypeId=LOCAL_FILE, createdByUserLogin=admin}
```

content.asImmutable().toString():
```
{ownerContentId=null, createdStamp=2021-01-19 13:57:12.593, contentId=10030, contentTypeId=DOCUMENT, 
description=null, mimeTypeId=image/png, childLeafCount=null, privilegeEnumId=null, dataSourceId=null, 
lastModifiedByUserLogin=admin, dataResourceId=10030, createdByUserLogin=admin, 
lastUpdatedStamp=2021-01-19 13:57:12.593, templateDataResourceId=null, lastModifiedDate=2021-01-19 13:57:12.593,
 instanceOfContentId=null, createdTxStamp=2021-01-19 13:57:09.983, lastUpdatedTxStamp=2021-01-19 13:57:09.983,
 serviceName=null, customMethodId=null, characterSetId=null, contentName=TestLogo.png, 
createdDate=2021-01-19 13:57:12.593, statusId=CTNT_PUBLISHED, localeString=null, 
decoratorContentId=null, childBranchCount=null}
```

partyContent.asImmutable().toString()
```
{fromDate=2021-01-19 13:57:13.014, lastUpdatedStamp=2021-01-19 13:57:13.014, partyContentTypeId=LGOIMGURL, 
createdTxStamp=2021-01-19 13:57:09.983, createdStamp=2021-01-19 13:57:13.014, contentId=10030, 
lastUpdatedTxStamp=2021-01-19 13:57:09.983, partyId=Company, thruDate=null}
```

I tried the following methods at calling PartyContentWrapper.getPartyContentAsText from CompanyHeader.groovy:

PartyContentWrapper.getPartyContentAsText(partyGroup, content.getString("contentId"), context.request, "html")
`"10030"`

PartyContentWrapper.getPartyContentAsText(partyGroup, content.getString("contentId"), context.request, "url")
`"10030"`

PartyContentWrapper.getPartyContentAsText(partyGroup, content.getString("contentId"), context.request, null)
`"10030"`

I'm not sure how else I should be using getPartyContentAsText as part of retrieving the binary content in order to convert to a data-uri.

dixitdeepak: After debugging code it fails to return object info due the check at DataResourceWorker.writeDataResourceText method
https://github.com/apache/ofbiz-framework/blob/trunk/applications/content/src/main/java/org/apache/ofbiz/content/data/DataResourceWorker.java#L927

`            if (dataResourceMimeTypeId == null || dataResourceMimeTypeId.startsWith("text")) {`

If we delete the dataResourceMimeTypeId field from DataResource, it will work

danwatford: Hi @dixitdeepak , 
After manually setting DataResource.mimeTypeId to null I was able to call:
```
PartyContentWrapper.getPartyContentAsText(partyGroup, content.getString("contentId"), context.request, "html")
```
and get a String in response of the form `%EF%BF%BDPNG%0D%0A%1A...` which looks like a textual representation of the PNG file contents that I could then appropriately convert to base64 and include as a data-uri which is then consumed by Apache FOP.

However, to do this we had to manually amend a database record which was created as a result of uploading party content via the UI. It is not clear to me how where the problem is. 

Is there an issue with how content is handled when uploaded to ofbiz?
Is there an issue which how content is rendered to text? Do we need a way to tell PartyContentRenderer/DataResourceWorker that we wish to render a binary file as a base64 encoded string? I'm not really familiar with the content handling mechanisms in ofbiz so am not able to advise a direction at the moment.

It seems these issues might take a while to resolve. Meanwhile this PR is a workaround to get company logos displayed correctly in PDF documents and I would rather not delay the workaround unless there is a problem with its implementation.

Would you object to me merging the PR as is?

dixitdeepak: I don't think there is way to get base64 from content/dataresouce worker classes,

applications/order/groovyScripts/order/CompanyHeader.groovy
```diff
@@ -135,17 +136,25 @@ if (!partyId) {
...
-    } else {
-        if (partyGroup?.logoImageUrl) {
-            logoImageUrl = partyGroup.logoImageUrl
+    GenericValue dataResource =
+            PartyContentWrapper.getFirstPartyContentByType(partyId, partyGroup, "LGOIMGURL", delegator)
+                    ?.getRelatedOne("Content", true)
+                    ?.getRelatedOne("DataResource", true)
+
+    if (dataResource) {
+        File logoFile = DataResourceWorker.getContentFile(dataResource.getString("dataResourceTypeId"),
```
dixitdeepak: I think here we should add following check to make sure we are getting base64 in case of file
`if (dataResourceTypeId.indexOf("_FILE") >= 0) {`

As content can be configured in different ways, and can have a complete url it self

I'd recommand to  add /content/control/stream?contentId in else part.

applications/order/groovyScripts/order/CompanyHeader.groovy
dixitdeepak: I won't object :)
added comments to add condition while getting the data it'd be good we think around it as well.

applications/order/groovyScripts/order/CompanyHeader.groovy
danwatford: I've added the resource type id check and the fall-back to the previously used content/control/stream url.

PR #270, opened 2021-02-10
Improved: Switch from jCenter to mavenCentral to handle the jCenter shutdown (OFBIZ-12171)

mbrohl:
This has to be applied in synch with the lpugins pull request with the same title.
https://github.com/apache/ofbiz-plugins/pull/53


mbrohl:
Strange, this worked on my side now, will check again...

mbrohl:
Ah, I see. This is a problem which is fixed in the ldap plugin but the fix is not available for this build because it is not committed to the official repo yet. Please check with both PR's for plugins and framework.

PR #283, opened 2021-03-02
OFBIZ-11683 Fixed: Humanres tree allows existing OU to be added again

SebastianEcomify:
Fixed: Humanres tree allows existing OU to be added again (OFBIZ-11683)

roleTypeId 'INTERNAL_ORGANIZATIO' is added for PartyRelationship in the
DemoData to prevent duplicate entries.

Thanks Pierre Smiths for reporting the issue

### DIFF
--- a/applications/datamodel/data/demo/AccountingDemoData.xml
+++ b/applications/datamodel/data/demo/AccountingDemoData.xml
@@ -1623,7 +1623,7 @@ under the License.
     <PartyRole partyId="MARKETING" roleTypeId="INTERNAL_ORGANIZATIO"/>
     <PartyRole partyId="MARKETING" roleTypeId="_NA_"/>
     <PartyStatus partyId="MARKETING" statusId="PARTY_ENABLED" statusDate="2001-01-01 12:00:00.0"/>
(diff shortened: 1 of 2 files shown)


JacquesLeRoux:
HJi Sebastian,

I tested and after adding the data I was able to add a "company" to "company" as it exists in trunk HEAD. Then you get an infitinte recursive tree that does not seem to fit to me.

Also please comment your change to be used in the description of the commit and of course to better understand what you are doing, TIA

PierreSmits:
Hi Sebastian, all,

We should not regard departments as internal organisations (we should reserve that for fiscal/legal entities, requiring their own CoAs). 

The roll-up for departments should be:
 <PartyRelationship partyIdFrom="Company" partyIdTo="MARKETING" partyRelationshipTypeId="GROUP_ROLLUP" roleTypeIdFrom="INTERNAL_ORGANIZATIO" roleTypeIdTo="DEPARTMENT" fromDate="2001-05-13 00:00:00.000"/>
 
 Best regards,
 Pierre

PierreSmits:
Hi,

Why are you treating departments of the internal organisation (partyID=company) as internal organisations as well? Should these have their own accounting preferences, as well as having their own gl schema?

Clearly, departments, teams, etc. should not be treated as internal organisations but rather as what they are intended to portray.

PR #284, opened 2021-03-05
 OFBIZ-12193 Fixed: Cant search ViewEntity with partial group-by fields

SebastianEcomify:
Fixed: Cant search ViewEntity with partial group-by fields (OFBIZ-12193)

For ViewEntities the SELECT statement can only contain function Fields
or the Fields need to be after GROUP_BY as well. The fix constricts the
select Fields for ViewEntities to functionFields and groupByFields.


JacquesLeRoux:
This has been already pushed, closing

PR #293, opened 2021-04-07
Improved: Have the ability to revoke (expire) multipe roles of a part…

enehlauri:
…y in one action (OFBIZ-5980)

Created a button to clear all roles in a party.

Thanks: laurakurkinen, venlavie PauHip for contributing

### DIFF
--- a/applications/party/config/PartyUiLabels.xml
+++ b/applications/party/config/PartyUiLabels.xml
@@ -11660,6 +11660,9 @@
         <value xml:lang="zh">关系</value>
         <value xml:lang="zh-TW">關連</value>
     </property>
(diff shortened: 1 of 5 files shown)


JacquesLeRoux:
Hi,

In OFBiz we prefer to expire than revoke things. The idea is that, unlike when expired where the thru date is set, when revoked the information is deleted hence lost. Please next time use rather https://issues.apache.org/jira/browse/OFBIZ-5980 to exchange, TIA

JacquesLeRoux:
Thanks for your work guys,

We have decided to apply the work done at https://issues.apache.org/jira/browse/OFBIZ-5980 (with explanation why there). Closing here.

PR #294, opened 2021-04-11
Implemented: Add new party while placing quotes (OFBIZ-7456)

Jontzii:
Introduce the option to create a new party group while placing the quotes in the system.

Credits: laslauk, jarirauhala, JesperNiemela, Jontzii


JacquesLeRoux:
Thanks guys!

PR #297, opened 2021-05-26
Improved: MacroFormRenderer refactoring of field group open and close elements (OFBIZ-12124)

danwatford:
Part of the OFBIZ-11456 MacroFormRenderer refactoring effort.

Rather than MacroFormRenderer producing and evaulating FTL strings, it now
uses RenderableFtlElementsBuilder to create RenderableFtlMacroCall objects
which are then passed to an FtlWriter for evaluation.


danwatford:
Merged

PR #303, opened 2021-06-14
Fixed: Adding tel protocol in CustomPermissivePolicy is not working (OFBIZ-12258)

wpaetzold:
With these changes, the default sanitizers rules are only loaded if there is no CustomPermissivePolicy


JacquesLeRoux:
Hi Wiebke,

I only reviewed (did not test), but this PR sounds ready to push to me

PR #307, opened 2021-07-02
Implemented: Add FtlWriter.processWithArgs (OFBIZ-12274)

dram:
When calling FTL macro from Java side is needed, instead of crafting a template string with carefully escaping. this patch try to pass parameters using an environment variable, and then invoke macro with "with_args" builtin. With this approach, it both makes cleaner code and free of fragile escaping.

This commit also changes MacroFormRenderer.makeHyperlinkString to make use of this new facility.


dram:
Closing, prefer to #314 by @danwatford

PR #308, opened 2021-07-03
Improved: Cleaned commented code from entity component (OFBIZ-8038)

ieugen:
Improved: Cleaned commented code from entity component (OFBIZ-8038)


ieugen:
Some comments might be reverted. 
They seem to be used for debugging.
I removed them to make them stand out for review.

framework/entity/src/main/java/org/apache/ofbiz/entity/model/ModelReader.java
```diff
@@ -225,8 +224,7 @@ private ModelEntity buildEntity(ResourceHandler entityResourceHandler, Element c
                             throw new GenericEntityConfException("Could not get document for " + entityResourceHandler.toString());
                         }
 
-                        // utilTimer.timerString("Before getDocumentElement in " +
```
JacquesLeRoux, line 228: Hi Eugen,

Do we really need this one?

framework/entity/src/main/java/org/apache/ofbiz/entity/util/EntityCrypto.java
```diff
@@ -110,16 +110,6 @@ public String encrypt(String keyName, EncryptMethod encryptMethod, Object obj) t
         }
     }
```
JacquesLeRoux: I think we need to keep this one

ieugen: Added back.

framework/entity/src/main/java/org/apache/ofbiz/entity/model/ModelReader.java
ieugen: I've kept one line comments that seem to be useful in debugging. 
They are not that many so have little impact.

ieugen:
Ready for merge.

PR #310, opened 2021-07-07
Improved: Clean commented code from ofbiz services

ieugen:
Improved: Clean commented code from ofbiz services

(OFBIZ-8040)


ieugen:
Ignored one line, non-code comments and non-code comments in general.

PR #317, opened 2021-07-28
Clojure engine for OFBiz

ieugen:
Implemented: Clojure engine for OFBiz

Explanation: Implemented clojure engine, test for the engine and service example

```
2021-07-28 23:11:06,453 |jsse-nio-8443-exec-4 |StandardClojureEngine         |I| Created Clojure engine.
2021-07-28 23:11:06,454 |jsse-nio-8443-exec-4 |StandardClojureEngine         |I| Call org.apache.ofbiz.common.clojure-test/echo-service 
```

### DIFF
--- a/framework/common/servicedef/services.xml
+++ b/framework/common/servicedef/services.xml
@@ -43,7 +43,11 @@ under the License.
 
     <service name="echoService" engine="java" validate="false"
             location="org.apache.ofbiz.common.CommonServices" invoke="echoService">
-        <description>Echos back all passed parameters</description>
+        <description>Echoes back all passed parameters</description>
+    </service>
+    <service name="echoClojureService" engine="clojure" validate="false"
+             location="org.apache.ofbiz.common.clojure-test" invoke="echo-service">
+        <description>Echoes back all passed parameters</description>
     </service>
     <service name="returnErrorService" engine="java" validate="false"
             location="org.apache.ofbiz.common.CommonServices" invoke="returnErrorService">
--- a/framework/common/servicedef/services_test.xml
+++ b/framework/common/servicedef/services_test.xml
@@ -47,6 +47,13 @@ under the License.
     <service name="testError" engine="java" export="true" validate="false" require-new-transaction="true" max-retry="1"
             location="org.apache.ofbiz.common.CommonServices" invoke="returnErrorService">
     </service>
+    <service name="testClojureSvc" engine="clojure" export="true" validate="false" require-new-transaction="true"
+             location="org.apache.ofbiz.common.clojure-test" invoke="test-clojure-svc">
+        <description>Test service</description>
+        <attribute name="defaultValue" type="Double" mode="IN" default-value="999.9999"/>
+        <attribute name="message" type="String" mode="IN" optional="true"/>
+        <attribute name="resp" type="String" mode="OUT"/>
+    </service>
     <!-- Because of the danger of Java deserialization when using RMI, the RMI component has been disabled in the default configuration of OFBiz.
          If you need RMI you just need to uncomment those places - See OFBIZ-6942 for details -->
     <!-- see serviceengine.xml to configure the rmi location alias -->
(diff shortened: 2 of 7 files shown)


framework/common/servicedef/services.xml
```diff
@@ -45,6 +45,10 @@ under the License.
             location="org.apache.ofbiz.common.CommonServices" invoke="echoService">
         <description>Echos back all passed parameters</description>
     </service>
+    <service name="echo-service" engine="clojure" validate="false"
```
nmalin: I suggest to use camel case for the name, like 'echoClojureService'

framework/common/servicedef/services_test.xml
```diff
@@ -47,6 +47,13 @@ under the License.
     <service name="testError" engine="java" export="true" validate="false" require-new-transaction="true" max-retry="1"
             location="org.apache.ofbiz.common.CommonServices" invoke="returnErrorService">
     </service>
+    <service name="test-clojure-svc" engine="clojure" export="true" validate="false" require-new-transaction="true"
```
nmalin: Same as previous :) : 'testClojureSvc'

framework/common/servicedef/services.xml
ieugen: is this a hard rule? 
echo-service is closer to clojure semantics. 
Clojure can have functions called like: "are-we>there-yet?". 
If it's not a requirement, I would prefer to keep it as is. 


Also, IMO OFBiz could use namespaces for services - it will help in grouping them based on functionality and will keep the names shorter:

party/xxxx
order/xxx
accounting/xxxx

ieugen: Namespaces could also help in people providing similar services under their own plugin / name without conflicting with existing services.

framework/common/servicedef/services_test.xml
ieugen: If it's not a hard rule I would prefer to keep them as is.

framework/common/servicedef/services.xml
nmalin: At the begining I would reponse that its a hard rule but after wound up the timeline I saw : 
 * a cleanning of bad service naming [1]
 * no information on it in OFBiz best pratice (only on databse) [2] , only on entity name
 * remember a discussion from Adrian Crum but I couldn't get my hands on it again

At this time all service follow this rule, so this question is interresting to what rule we need to follow, because the service engine is a layer to abstract the language so why follow a rule link to a language but in other way the service is dedicate to test a engine so linked to this language.

My feel would be follow the layered practice and invisible the language behind (exclude the service to test an engine :sweat_smile: ) .

Concerning the namespace, clearly is a subject but i'm not sure to involve it here with blur the main purpose. I suggest to share on this part after directly on dev list :)

[1] https://svn.apache.org/viewvc?view=revision&revision=1369948
[2] https://cwiki.apache.org/confluence/display/OFBIZ/Best+Practices+Guide

framework/common/servicedef/services_test.xml
ieugen: Done.

framework/common/servicedef/services.xml
ieugen: Renamed. 
We can approach this issue at a later time. 
I'll make a proposal in the next future to make OFBiz composable and leaner.

ieugen:
I think it's ready @nmalin .

nmalin:
Yes I confirm, I will squash and push.
Do you have a linked jira issue for the backlog ? I didn't found it.

To prevent a squash form my part, peraphs you can do it with [a commit message that follow the template](https://cwiki.apache.org/confluence/display/OFBIZ/OFBiz+commit+message+template) .

ieugen:
@nmalin : Added jira issue and message + squashed commits. https://issues.apache.org/jira/browse/OFBIZ-12302

nmalin:
Works for me thank Eugeu !

PR #323, opened 2021-10-28
Improved: partyRole entity (OFBIZ-5980)

PierreSmits:
added fields 'fromDate' and 'thruDate' to allow for party role expiration


JacquesLeRoux (changes requested):
I don't think we want to remove the comment



 in HumanresDemoData.xml

PierreSmits:
Hi @JacquesLeRoux,

Regularly you advice other contributors to leave the comment on the associated ticket. Please stick to that advice too.

JacquesLeRoux:
In this case it's because I made a review here.

PR #327, opened 2021-11-02
Fixed:  Quote item quantity field not visable (OFBIZ-12357)

PierreSmits:
removed: display attribute

Thanks: Justine Nowak for opening the ticket.


mbrohl:
Hi Pierre,
this PR with 2290 commits included does not seem to be right - can you please check?

PierreSmits:
Hi Michael,

Thanks for the observation. This should not have been for such a small change. Closing

PR #345, opened 2021-11-16
 Improved: create invoice service (OFBIZ-12387)

PierreSmits:
added:
InvoiceServices.groovy: def createInvoice
- the added service also checks and ensures that the user has the appropriate PartyRole
- the added service also creates the InvoiceRole record for the user
AccountingSeedData.xml
- RoleType definitions for InvoiceRole

modified:
services_invoice.xml, createInvoice: engine attribute and location attribute

removed:
InvoiceServices.xml: simple-method createInvoice

### DIFF
--- a/applications/accounting/groovyScripts/invoice/InvoiceServices.groovy
+++ b/applications/accounting/groovyScripts/invoice/InvoiceServices.groovy
@@ -25,6 +25,60 @@ import org.apache.ofbiz.service.ServiceUtil
 
 import java.sql.Timestamp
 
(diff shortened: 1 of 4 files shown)


mbrohl:
The changes are not only a conversion to Groovy but also introduce changes to the service business logic.
The conversion should be a separate PR, the changes should be in another PR with some explanation why they are necessary.

JacquesLeRoux:
I agree with Michael, explanations are less about what you did, but more about why you did them.

PR #351, opened 2021-11-19
Improved: OFBiz audit/investigation (OFBIZ-12391)

PierreSmits:
modified:
ModelEntity.java - adding fields for auditing/investigating record creation and modification by default
to entities when OFBiz is initialised:
 * createdDate
 * createdByUserLogin
 * lastModifiedDate
 * lastModifiedByUserLogin

### DIFF
--- a/applications/datamodel/entitydef/accounting-entitymodel.xml
+++ b/applications/datamodel/entitydef/accounting-entitymodel.xml
@@ -1788,10 +1788,6 @@ under the License.
       <field name="receiptId" type="id"></field>
       <field name="workEffortId" type="id"></field>
       <field name="theirAcctgTransId" type="id-long"></field>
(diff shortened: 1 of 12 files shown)


mbrohl:
AFAIK this can all be configured in the entity model as I mentioned in another PR some days ago.
Why is this change necessary?

PierreSmits:
Thank you for your feedback, Michael,

I am confident that you mentioned something somewhere about the aspect that something could be done somewhere else in OFBiz, but without actual reference to where this happened it hard to check whether it is pertaining to changes under this PR and/or related to the ticket associated. It could even be imaginary (not saying that it is).

So please provide the reference (preferably as a link), and if those comment there/then do not explicitly relate to the changes under this PR please also provide reasoning why it should.

It will surely help other contributors to form an opinion on whether having this in the code base, or not.

PR #356, opened 2021-11-25
Improved: redundancy removal (OFBIZ-12404)

PierreSmits:
modified: InvoiceScreens.xml
removed redundant property-maps for Party and Product labels from InvoiceOverview

### DIFF
--- a/applications/accounting/widget/InvoiceScreens.xml
+++ b/applications/accounting/widget/InvoiceScreens.xml
@@ -182,8 +182,6 @@ under the License.
     <screen name="invoiceOverview">
         <section>
             <actions>
-                <property-map resource="PartyUiLabels" map-name="uiLabelMap" global="true"/>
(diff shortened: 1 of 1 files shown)


JacquesLeRoux:
Hi Pierre,

Even if it obvious for you (once done), it would have been great if you had added that these lines duplicated those in main-decorator (I guess)

PierreSmits:
Thanks Jacques, for your constructive comment. Context is addressed in the ticket, which will be seen by existing and potential adopters when the release notes are generated for a new release. Context provided in the commit is not referenced there.

JacquesLeRoux:
Right, but it would not have hurt to have it here too :unamused:

PierreSmits:
Jacques, 
Please consider that this may trigger someone to cry wolf again, as we have seen in recent days. Let it go. For harmony (and happy collaborating) sake.

PR #363, opened 2021-11-28
Improved: VIEW permissions and invoice attributes (OFBIZ-12417)

PierreSmits:
Though there is an entity InvoiceAttribute, this cannot be accessed by user with VIEW permissions when accessing an invoice.

Modified:
InvoiceScreens.xml
- added screen InvoiceAttributes to be available for users with VIEW permissions
InvoiceForms.xml
- added grid InvoiceAttributes to be used by screen InvoiceAttributes
AccountingMenus.xml
- added menu-item invoiceAttributes to InvoiceTabBar
controller.xml
- added request-map and view-map to access screen InvoiceAttributes

### DIFF
--- a/applications/accounting/config/AccountingUiLabels.xml
+++ b/applications/accounting/config/AccountingUiLabels.xml
@@ -24163,6 +24163,10 @@
         <value xml:lang="zh">发票总览</value>
         <value xml:lang="zh-TW">發票總覽</value>
     </property>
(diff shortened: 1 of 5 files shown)


JacquesLeRoux:
Yes, why not?

PierreSmits:
Why? 

I guess we both can think of a lot of reasons. But we should explore that on dev.

Thanks Jacques.

PR #381, opened 2021-12-03
Exploratory Introduced framework, applications, themes projects (OFBIZ-3500)

ieugen:
Another exploratory patch - as a PR for easy review.

* First steps into splitting OFBiz into componets
* 5 tests don't pass in framework
* ofbiz distribution builds
* App fails to start because of component loading logic

Right now I am stuck and don't know how to proceed with loading components. 

Loading components via component-load.xml and ofbiz-comonent.xml is a mix of file path and classpath. 

Currently OFBIz does not know how to load components packaged as a jar.
Components are REQUIRED to be on disk as much as I can tell. 
This is also referenced in ofbiz-component.xml and component-load.xml files via things like:
```   
    <classpath type="dir" location="config"/>
```
and the fact they try to load from filesystem.

IMO Java is not meant to be run from source files and OFBiz tries to do just that. 

There is no way - to my knowledge - to package a component as a single archive and distribute it. 

IMO, fixing component packaging and loading is a key part of making OFBiz modular. 

I will post this on the mailing list to get some feedback. 
For code related feedback, please comment on this PR.


ieugen:
Deprecated in favor of https://github.com/apache/ofbiz-framework/pull/678 .

PR #384, opened 2021-12-04
Improved: VIEW permissions Commission Run (OFBIZ-12433)

PierreSmits:
Currently, when using a userId with only VIEW permission, the Commission Run trigger (menuItem) presents a screen to search and find invoices applicable for the underlying service of creating commission invoices.
On the result shown, the user has access to triggers (functionalities) reserved for users with CREATE/UPDATE permissions.
See (test with): https://demo-trunk.ofbiz.apache.org/accounting/control/CommissionRun and from date= 12/1/1999 12:00:00 AM

Modified:
InvoiceScreens.xml (in AP) - screen CommissionRun
- restructured the screen to show result based on permissions
- added reference to template for users with VIEW permissions

Added:
CommissionRunSearchResult.ftl template for users with VIEW permissions

### DIFF
--- a/applications/accounting/template/invoice/CommissionRunSearchResult.ftl
+++ b/applications/accounting/template/invoice/CommissionRunSearchResult.ftl
@@ -1,0 +1,66 @@
+<#--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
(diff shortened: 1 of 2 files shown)


JacquesLeRoux:
Hi Pierre,

Why using Freemarker rather than Widgets?

PierreSmits:
Good afternoon Jacques,

For several reasons:
1. There is already a template file for users with CREATE/UPDATE permissions, and to keep contribution experience similar to that
2. This ticket is not about conversions of template code to xml code, but about enhancing the trustworthiness of OFBiz from the viewpoint of the user.

JacquesLeRoux:
I was only interested in point 1, but thanks

PierreSmits:
Then you can regard point 2 as a bonus. :)

PR #391, opened 2021-12-06
Improved: AccountingUiLabels clean-up (OFBIZ-9352)

PierreSmits:
Removing unused labels from AccountingUiLabels.xml

Checked screens, forms, templates, scripts and java code

### DIFF
--- a/applications/accounting/config/AccountingUiLabels.xml
+++ b/applications/accounting/config/AccountingUiLabels.xml
@@ -2050,44 +2050,6 @@
         <value xml:lang="zh">账单账户编号</value>
         <value xml:lang="zh-TW">帳單帳戶識別</value>
     </property>
-    <property key="AccountingBillingAccountLessRequested">
-        <value xml:lang="ar">حساب الفوترة  [${billingAccountId}]لديه ${billingAccountApplyAvailable?currency(${isoCode})} للتطبيق لكن ${amountApplied?currency(${isoCode})}مطلوبة</value>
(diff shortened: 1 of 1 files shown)


JacquesLeRoux:
Pierre,

Today I asked myself (again): "Some efforts have been done to produce those. Are we sure we want to remove them? This does not break anything and can still be used in projects that could be updated and missing these labels." 

I then checked, we had this discussion many time already (mostly, in order of appearance):
https://markmail.org/message/so4gyq6ph4udh6cm (2016)
https://issues.apache.org/jira/browse/OFBIZ-8154 (2016)
https://issues.apache.org/jira/browse/OFBIZ-9352 (2017)
https://issues.apache.org/jira/browse/OFBIZ-10565 (2018 umbrella task)
https://markmail.org/message/7kckrtpzw5r7fvkl (2020)

After that I decided to apply what we discussed before: check if it's not used anywhere, then remove. I'm still not sure it's the best decision, but it's the one we decided on.

PierreSmits:
Hi Jacques,

Thank you for taking the time to look back, and see why something came about to be. 

These things, like so many others in the codebase, came to be in there (and this is not to lay blame, just an observation), because at the time it seemed appropriate to bring them in and intended - most probably - to be followed up by the contributor in a next ticket. Or maybe it was part of an implementation at a user's site, but the rest was not allowed to go into the project's codebase or forgotten. But you know, life gets in the way or career takes a turn. 

And the good thing is, nothing is ever lost. So we can clean up when not used together with something else in OFBiz. And if in future some other contributor can make a case for an improvement and bring a past included and removed code element back, good for for it.

JacquesLeRoux:
Yes, we can always revert (easy with labels, not much changes now).

PR #392, opened 2021-12-07
Improved: Have configuration options for accounting (OFBIZ-6189)

PierreSmits:
added: SystemProperty records in AccountingSystemPropertyData.xml
added: reference of AccountingSystemPropertyData.xml to load at seed loading moment


JacquesLeRoux:
I'll now adapt the commit comment when there are several commits...

PR #393, opened 2021-12-07
Improved: VIEW permissions FinAccount roles (OFBIZ-12435)

PierreSmits:
Currently, a user with only 'VIEW' permissions, as demonstrated in trunk demo with userId = auditor, accessing the Financial Account Roles screen  sees editable fields and/or triggers (to requests) reserved for users with 'CREATE' or 'UPDATE' permissions.

See (test with): https://demo-trunk.ofbiz.apache.org/accounting/control/EditFinAccount?finAccountId=ABN_CHECKING

Modified:
FinAccountScreens.xml - restructured EditFinAccount
FinAccountForms.xml - added FinAccountRoles form, restructured ListFinAccountRoles

### DIFF
--- a/applications/accounting/widget/FinAccountForms.xml
+++ b/applications/accounting/widget/FinAccountForms.xml
@@ -176,8 +176,18 @@ under the License.
         <field name="availableBalance"><display type="accounting-number"/></field>
     </form>
 
-    <grid name="ListFinAccountRoles" list-name="finAccountRoles"  target="updateFinAccountRole" separate-columns="true"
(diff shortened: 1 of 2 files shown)


JacquesLeRoux:
Pierre,

I guess you will continue to disable for auditor (or alike) the screens accessible from this screen? As it would not make sense to disable the menus themselves, right?

PierreSmits:
Hi Jacques,

Thanks for the question. 
How do you mean 'disable the menus'? As in the whole menu for a given screen? Or multiple screens related to same object?

How would users then go from one sub element to another given the same object?

JacquesLeRoux:
Forget it, I just meant that the screens accessible from the sub-menus should still be accessible, nothing new.

PR #398, opened 2021-12-08
Improved: Have configuration options for Content (OFBIZ-6670)

PierreSmits:
The configuration of an OFBiz is determined by 2 aspects:
configuration settings required to successfully start the implementation
configuration settings required to have component functions work properly
The latter involves default choices for various functions to cater to the usability of business processes (and therefore screen/forms and/or screens/templates.
Several of these configuration settings are already loaded into the SystemProperty table in the underlying RDBMS, to enable adjustments of the systemPropertyValue while the implementation is in use. While it is not expected that these settings change often, it is possible that they are required to be changed during the lifespan of the company/organisation.
However, it should not be required to go through a deployment process of regarding such changed configuration settings in the OFBiz production instance (which most - if not all - companies/organisations require when files are changes).
Having all configuration settings of components in the SystemProperty table enhances the usability of OFBiz allowing system administrators to adjust the value(s) of system property records while the instance keeps running.

### DIFF
--- a/applications/content/data/ContentSystemPropertyData.xml
+++ b/applications/content/data/ContentSystemPropertyData.xml
@@ -22,6 +22,10 @@ under the License.
     systemPropertyValue="CONTENT_ROOT"
      description="content id regarding base content owner"/>
 
+    <SystemProperty systemResourceId="content" systemPropertyId="img.request" 
+    systemPropertyValue="/content/control/img/"
+     description="uri path for image requests"/>
+
     <SystemProperty systemResourceId="content" systemPropertyId="img.request.param.name" 
     systemPropertyValue="imgId"
      description="request param for images"/>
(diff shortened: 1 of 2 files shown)


JacquesLeRoux (changes requested):
Hi Pierre, 

Why not img.request and stream.buffersize?

PierreSmits:
Hi Jacques,

My apologies. I must have missed those two. I will amend.

Thanks for the observation and the review.

mbrohl:
@JacquesLeRoux Can we please try to clear open questions from other contributors towards those implementations?
There already is an open question from @pythys in the Jira which is not answered.

Also, I think it is not a good idea to load this data as seed data. There is a danger that configured settings (through properties) are secretly overriden by this. In the past, we had very few default settings which are automatically loaded in the SystemProperty entity.

Having the data available is ok with me, but please do not automatically load it.

JacquesLeRoux:
Hi @mbrohl,

I already answered in the Jira, please answer there, TIA

PR #410, opened 2021-12-13
Improved: VIEW permissions FinAccount mutations (OFBIZ-12451)

PierreSmits:
Currently, a user with only 'VIEW' permissions, as demonstrated in trunk demo with userId = auditor, accessing the Financial Account mutations screen  sees editable fields and/or triggers (to requests) reserved for users with 'CREATE' or 'UPDATE' permissions.
See (test with):
https://demo-trunk.ofbiz.apache.org/accounting/control/EditFinAccount?finAccountId=ABN_CHECKING
https://demo-trunk.ofbiz.apache.org/accounting/control/FindPaymentsForDepositOrWithdraw

Modified:
- FinAccountScreens.xml: work with permissions, added ref to mutations template for user with VIEW permission
Added:
- Mutations.ftl template

### DIFF
--- a/applications/accounting/template/finaccount/Mutations.ftl
+++ b/applications/accounting/template/finaccount/Mutations.ftl
@@ -1,0 +1,45 @@
+<#--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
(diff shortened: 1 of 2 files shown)


JacquesLeRoux:
Pierre,

Why are some elements not present in the view only?
Fin Account Name
From Date

PierreSmits:
Hi Jacques,

It is true that the underlying entity used to display the line items has more elements than is being displayed with the change. This is controlled by the DepositWithdrawPayments.groovy script, which pulls data from PaymentAndTypePartyNameView. 

Neither Fin Account Name, nor From Date are part of that dataset. 

While the script potentially could provide more data elements from the source, the original implementation doesn't show that (see with login = admin). So that has been followed here. We can leave that at the hands of downstream implementers.

PR #412, opened 2021-12-13
Improved: Rename invoiceApplications menu-item label to Payments (OFBIZ-12405)

PierreSmits:
With commit 4aa2d71 ambiguity was introduced into the label AccountingApplications and subsequently
applied to the menu-item for payments in the menu for invoices. fixing the applicable menu-item

Modified: AccountingMenus.xml
- label reference in the title for the menu-item for invoice payments
addtionally: setting the labels for titles of 'overview' menu-items to that in CommonUiLabels.xml, for:
- budget overview
- invoice overview
- payment overview
- payment group overview

### DIFF
--- a/applications/accounting/widget/AccountingMenus.xml
+++ b/applications/accounting/widget/AccountingMenus.xml
@@ -117,7 +117,7 @@ under the License.
         <menu-item name="find" title="${uiLabelMap.CommonFind}">
             <link target="findInvoices"/>
         </menu-item>
-        <menu-item name="invoiceOverview" title="${uiLabelMap.AccountingInvoiceOverview}">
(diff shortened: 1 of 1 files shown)


mbrohl:
As we already discussed in, for example, in https://issues.apache.org/jira/browse/OFBIZ-8109 and https://issues.apache.org/jira/browse/OFBIZ-8056, those changes will cut the flexibility for naming and translations.

If a user wants to name those titles differently, for whatever reason, he has to into re-introduce the labels back again to reach this goal. We should keep the flexibility to name each application's UI elements independently of others even if we might have them named the same ootb.

mbrohl:
Closed, won't do.

PR #416, opened 2021-12-15
Improved: Cybersource - configuration (OFBIZ-12459)

PierreSmits:
Move 3rd party Fintech integration configuration for Cybersource from initialisation configuration to runtime configuration.

Modified:
- payment.properties: removed Cybersource configuration aspects
- AccountingSystemPropertiesData.xml: Cybersource configuration aspects
- icsPaymentServices.java: switched from getPaymentGatewayConfigValue to EntityUtilProperties.getPropertyValue for Cybersource configuration aspects

### DIFF
--- a/applications/accounting/config/payment.properties
+++ b/applications/accounting/config/payment.properties
@@ -64,67 +64,6 @@ payment.giftcert.reload.survey.pinNumber=giftCardPin
 payment.giftcert.type=ofbiz
 
 ############################################
-# CyberSource Configuration
-############################################
-
-# your merchant Id
-payment.cybersource.merchantID=[MerchantID]
-
-# target API version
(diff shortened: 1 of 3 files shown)


mbrohl:
This also ignores the objection to focus on SystemProperty only.
Removing the property file based configuration and replacing it with SystemProperty forces the user to use SystemProperty which is not reasonable.
This should not go into the code base as-is.

mbrohl:
This should not go into the code base as is. Please leave the established mechanism/properties as is and only add SytemPropertyData as optional load data.

JacquesLeRoux:
We recently agreed about that. Anyway this branch has conflicts that I don't want to handle, closing

PR #423, opened 2021-12-19
Improved:  List and Grid (OFBIZ-11345)

PierreSmits:
According to the definition in widget-form.xsd the use of a combination of a form with type="list" is deprecated in favour of a grid.
Refactor various list forms into grids.
Refactor various list form references in screens.

Modified:
SubscriptionScreens.xml: from form ref to grid ref , additional cleanup
SubscriptionForms.xml: from form definition with list ref to grid definition with list ref, additional clean-up

### DIFF
--- a/applications/product/widget/catalog/SubscriptionForms.xml
+++ b/applications/product/widget/catalog/SubscriptionForms.xml
@@ -25,14 +25,6 @@ under the License.
     <form name="FindSubscription" default-map-name="subscription" target="FindSubscription" type="single"
         header-row-style="header-row" default-table-style="basic-table">
         <auto-fields-entity entity-name="Subscription" default-field-type="find"/>
-
-        <field name="subscriptionResourceId">
-            <drop-down allow-empty="true">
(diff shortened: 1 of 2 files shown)


JacquesLeRoux:
Hi Pierre,

Could you please explain why you introduced unrelated changes eg in FindSubscription, ListFindSubscription? I guess it's the clean-up you speak about? Could this not have been part of another PR?

PierreSmits:
Hi Jacques,

Thanks for the review. 
The short answer on all three questions is 'Yes'.

JacquesLeRoux:
Then could you please split this part in another PR?

PierreSmits:
You talk about the labels added?

JacquesLeRoux:
Actually I was unable of locally patching all and was a bit tired. Now I see things better and I have only one question: why removing <sub-hyperlink from communicationEventId  lookup? I checked there are 21 other cases like that in code.

JacquesLeRoux:
Hi Pierre,

I'll push from here because GH merges it easily. I'll then make some modification after pulling locally

PR #425, opened 2021-12-20
(OFBIZ-10093) Replace elrte with Trumbowyg

bjugl:
Improved: Replaces Elrte with Trumbowyg

(OFBIZ-10093)

Thanks: 
Jacques for inspiration and guidance, Aditya Sharma for sharing thoughts and visions and Ingo Richter for finding this gem.


mbrohl:
Hi Benjamin, 
to avoid those merge commits you better do a rebase instead of a merge to update your feature branch.
Can you please provide a clean PR? Thanks!

JacquesLeRoux:
Hi Benjamin,

I'll review the CodeQL scanning results. There are often false or benign alerts...

JacquesLeRoux:
It's all 3rd lib issues so hard to fix w/o losing contact with their upgrades. I'll still check...

JacquesLeRoux:
I did not review yet. I suggest we consider to use gradle-npm to load the lib.

bjugl:
> I did not review yet. I suggest we consider to use gradle-npm to load the lib.

Lib is now loaded by gradle-npm.

Trumbowyg uses its own css. Since some classes have attributes that are overwritten in themes, they have to be supplemented with !important tags in two places. 

Thus we have to have an overwrite css. I moved it to "common/css/ ".
I would like to document this somewhere, because it will propably cause some issues, when the version of trumbowyg changes. Since comments in the package.json are not an option, where would you recommend to put such a comment?

JacquesLeRoux:
Thanks Benjamin,

Following [this advice](https://www.freecodecamp.org/news/json-comment-example-how-to-comment-in-json-files/), maybe as a data? To be checked, not sure it has no side-effects, I doubt it'd have...

PierreSmits:
@bjugl 

You could consider doing a page in Confluence (it has version control), with a reference (to that) in the appropriate code file.

JacquesLeRoux:
@PierreSmits 

The problem Pierre is that there is not really an " appropriate code file"

JacquesLeRoux:
@bjugl ,

I'm no longer able to review the js CodeQL analysis (because of npm use). There were no real issues. I saw for instance that they used things like `\.` in regexp that are of no use in js since you actually need `\\.` for escaping a dot in js. But anyway it was not a security issue, just that it's the same to do `\.min` than `.min`. We could have reported the other issues (like 4 or 5) but it's no longer available. So +1 to push.

applications/content/widget/CommonScreens.xml
```diff
@@ -33,6 +33,7 @@ under the License.
                 <set field="layoutSettings.companyName" from-field="uiLabelMap.ContentCompanyName" global="true"/>
                 <set field="layoutSettings.companySubtitle" from-field="uiLabelMap.ContentCompanySubtitle" global="true"/>
                 <set field="layoutSettings.styleSheets[]" value="/contentimages/css/contentForum.css" global="true"/>
+                <set field="layoutSettings.styleSheets[]" value="/contentimages/css/contentForum.css" global="true"/>
```
adityasharma7: Hello @bjugl 
Thanks for your contribution!
It seems you missed to updated the css path.

themes/common-theme/webapp/common/css/trumbowygCustom.css
```diff
@@ -0,0 +1,642 @@
+/**
```
adityasharma7: @bjugl 
Is there any custom styling we have added overriding the default styling of trumbowyg. If not, we could directly use the css files available in ui folder. WDYT?

adityasharma7 (changes requested):

bjugl:
> I'm no longer able to review the js CodeQL analysis (because of npm use). There were no real issues. I saw for instance that they used things like `\.` in regexp that are of no use un js since you actually need `\\.` for escaping a dot in js. But anyway it was not a security issue, just that it's the same to do `\.min` than `.min`. We could have reported the other issues (like 4 or 5) o but it's no longer available. So +1 to push.

All the Warnings were in the Plugins, of which we - as of now - only use one. There was one possible Code-Injection in the CleanPastbin Plugin, because script-tags were not filtered within the logic. This we can ignore, because I set trumbowyg up to not accept script tags anyhow (We do also sanitize serversided again, right?).

applications/content/widget/CommonScreens.xml
bjugl: This seems wired, I will check.

themes/common-theme/webapp/common/css/trumbowygCustom.css
bjugl: We need the custom css, because Helveticus Themes override the color of at least the Icon that switches to code view and Rainbow Stone does silly things to the outlines of the boxes, due to margins defined on all div containers. 
Those are the two changes I revered two, when I asked about the proper way to overwrite earlier in this PR.

applications/content/widget/CommonScreens.xml
bjugl: That seems to be a remnant of an early trial and error phase, that I did not clean up properly, sry.
I decided, to not put them here, but rather into "themes/common-theme/widget/CommonScreens.xml" in "GlobalActions". This way, everything is neatly in common-theme, apart from the WebSiteCMSNav.ftl. I would just clean up this extra line, or do you suggest another way of implementation?

JacquesLeRoux:
>We do also sanitize serversided again, right?

Yes inputs are sanitised

themes/common-theme/webapp/common/css/trumbowygCustom.css
adityasharma7: I think we should use the default trumbowyg css as is and the changes we want should go in custom css files like we did for Select2
https://github.com/apache/ofbiz-framework/blob/trunk/themes/rainbowstone/webapp/rainbowstone/style.css#L1919

This way we will keep track of custom changes easily. WDYT?

applications/content/widget/CommonScreens.xml
adityasharma7: I think adding to common-theme seems to be the right approach, you have taken. Thanks for the explanation!

adityasharma7:
> @bjugl ,
> 
> I'm no longer able to review the js CodeQL analysis (because of npm use). There were no real issues. I saw for instance that they used things like `\.` in regexp that are of no use in js since you actually need `\\.` for escaping a dot in js. But anyway it was not a security issue, just that it's the same to do `\.min` than `.min`. We could have reported the other issues (like 4 or 5) but it's no longer available. So +1 to push.

I think we can use `npm audit`
https://docs.npmjs.com/cli/v8/commands/npm-audit

JacquesLeRoux:
Hi @adityasharma7,

Is that not already done when adding a new package? For instance adding `"trumbowyg": "^2.25.1"` we get

```
> Task :npmInstall
added 1 package from 1 contributor and audited 7 packages in 2.083s
found 0 vulnerabilities
```

This said it seems npm audit has not the same power than CodeQL (or it finds less false positives :smile: )

bjugl:
Cleaned up the CommonScreens.xml in applications and reconfigured, so that the trumbowyg.min.css is now used. styles.css in themes override this, where necessary.

adityasharma7:
> Hi @adityasharma7,
> 
> Is that not already done when adding a new package? For instance adding `"trumbowyg": "^2.25.1"` we get
> 
> ```
> > Task :npmInstall
> added 1 package from 1 contributor and audited 7 packages in 2.083s
> found 0 vulnerabilities
> ```
> 
> This said it seems npm audit has not the same power than CodeQL (or it finds less false positives 😄 )


It seems 😄. There may be warnings for the reported CVEs, not the entire code scan

themes/rainbowstone/webapp/rainbowstone/style.css
```diff
@@ -1927,4 +1927,12 @@ html > /**/ body .jstree-default a {
 /* The custom CSS to show mouse pointer on links using featherlight plugin  */
 a[data-featherlight] {
     cursor: pointer;
-}
\ No newline at end of file
+}
+/*The custom css for Trumbowyg*/
+.trumbowyg-button-pane {
```
adityasharma7, line 1933: There seems an indentation issue in css in both the files. Could you please fix that?
Rest looks good to me. Thanks for your updates!

adityasharma7 (changes requested):

themes/rainbowstone/webapp/rainbowstone/style.css
```diff
@@ -1927,4 +1927,12 @@ html > /**/ body .jstree-default a {
...
+    margin: 0 !important; }
```
adityasharma7:
```suggestion
    margin: 0 !important;
}
```

themes/rainbowstone/webapp/rainbowstone/style.css
```diff
@@ -1927,4 +1927,12 @@ html > /**/ body .jstree-default a {
...
+  .trumbowyg-button-pane .trumbowyg-button-group {
+    display: inline-block; 
+    margin: 0 !important; }
```
adityasharma7:
```suggestion
    margin: 0 !important;
}
```

themes/rainbowstone/webapp/rainbowstone/style.css
```diff
@@ -1927,4 +1927,12 @@ html > /**/ body .jstree-default a {
...
+    .trumbowyg-button-pane button {
+        color: #000; }
```
adityasharma7:
```suggestion
        color: #000;
}
```

themes/helveticus/webapp/helveticus/style.css
```diff
@@ -0,0 +1,30 @@
...
+/* This CSS file is used for the Helveticus visual theme. */
+/* ----------------------------------------------------- */
+
+/* This theme uses a scalable, floating layout - so most
+   sizes are expressed in EMs. PX sizes are used only for
+   drop shadows and positioning around graphic images. */
+
+/*The custom css for Trumbowyg*/
+.trumbowyg-button-pane button {
+    color: #000; }
```
adityasharma7:
```suggestion
    color: #000;
}
```

adityasharma7 (changes requested):

adityasharma7 (approved):

adityasharma7:
Thanks @bjugl for your contribution!

PR #446, opened 2022-01-13
Improved: Config Screen - VIEW permission (OFBIZ-12493)

PierreSmits:
Currently, a user with only 'VIEW' permissions, as demonstrated in trunk demo with userId = auditor, accessing the config item screen, sees editable fields and/or triggers (to requests) reserved for users with 'CREATE' or 'UPDATE' permissions.
See (test with): https://localhost:8443/catalog/control/EditProductConfigItem?configItemId=SC00000

modified:
- ConfigScreens.xml - restructured screen EditProductConfigItem to work with permissions
- ConfigForms.xml - added form ProductConfigItem for users with VIEW permission


PierreSmits:
The failure of the Java CI with Gradle / build (pull_request) Failing after 2m — build is due to a hiccup in the action.

PR #451, opened 2022-01-15
Improved: Quote - Action Triggers, Permissions & ux (OFBIZ-12499)

PierreSmits:
Currently, as can be seen in https://localhost:8443/ordermgr/control/ViewQuote?quoteId=CQ0001, action triggers (for users with CREATE/UPDATE permissions) are located within screens instead of the action menu for quotes. Also triggers can be seen and invoked by users with VIEW permissions.

modified:
- OrderMenus.xml
added inQuoteSubTabBar menu-items for users with CREATE/UPDATE permissions (move from various quote screens), additional cleanup
- QuoteScreens.xml
added QuoteSubTabBar to CommonQuoteDecorator
moved quote label to above menus in CommonQuoteDecorator
removed QuoteSubTabBar from various quote screens
removed action triggers from various quote screens
additional cleanup, label harmonisation and ux consistency
- QuoteWorkEffortScreens.xml
removed action triggers from various quote screens
additional cleanup, label harmonisation and ux consistency

### DIFF
--- a/applications/order/widget/ordermgr/OrderMenus.xml
+++ b/applications/order/widget/ordermgr/OrderMenus.xml
@@ -360,7 +360,7 @@ under the License.
                 <parameter param-name="quoteId" from-field="quote.quoteId"/>
             </link>
         </menu-item>
-        <menu-item name="NewItem" title="${uiLabelMap.CommonNew} ${uiLabelMap.CommonItem}">
+        <menu-item name="NewItem" title="${uiLabelMap.OrderCreateOrderQuoteItem}">
             <condition>
                 <and>
                     <or>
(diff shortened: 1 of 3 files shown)


JacquesLeRoux:
Hi Pierre,

I don't know if it's planned, what about "Duplicate Quote" for an auditor?

PierreSmits:
Hi Jacques,

As you can get from the changes in this implementation, the PR and the ticket addressed the triggers that belonged in an  Action Menu intended for users with CREATE and/or UPDATE permissions, not what users with ViEW permissions should or should not see vis-a-vis screens/forms/grids, etc. 

Hence, the 'Duplicate Quote' and the like are not addressed. But intended to be addressed at a later stage through a different ticket and PR. And by whomever is compelled to work that issue.

JacquesLeRoux:
Hi Pierre,

Thanks for the confirmation

PR #465, opened 2022-01-21
Improved: dutch labels (OFBIZ-10363)

PierreSmits:
Improved: Dutch Labels (OFBIZ-10363)

modified: HumanResUiLabels.xml
Added dutch labels for Skill(s)


danwatford:
Checked with Google Translate.

PR #466, opened 2022-01-21
Improved: List and Grid (OFBIZ-11345)

PierreSmits:
According to the definition in widget-form.xsd the
use of a combination of a form with type="list" is
deprecated in favour of a grid.
Refactor various list forms into grids.
Refactor various list form references in screens.

modified:
RecruitmentForms.xml: from form definition with list
ref to grid definition with list ref, additional clean-up
Recruitmentcreens.xml: from form ref to grid ref , additional cleanup

### DIFF
--- a/applications/humanres/widget/RecruitmentScreens.xml
+++ b/applications/humanres/widget/RecruitmentScreens.xml
@@ -20,7 +20,7 @@ under the License.
 <screens xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
     xmlns="http://ofbiz.apache.org/Widget-Screen" xsi:schemaLocation="http://ofbiz.apache.org/Widget-Screen http://ofbiz.apache.org/dtds/widget-screen.xsd">
     <screen name="FindJobRequisitions">
-        <section>             
+        <section>
             <actions>   
                 <set field="titleProperty" value="PageTitleFindJobRequisition"/>
                 <set field="tabButtonItem" value="JobRequisition"/>
@@ -54,7 +54,7 @@ under the License.
                                         <include-form name="FindJobRequisitions" location="component://humanres/widget/forms/RecruitmentForms.xml"/>
                                     </decorator-section>
                                     <decorator-section name="search-results">
-                                        <include-form name="ListJobRequisitions" location="component://humanres/widget/forms/RecruitmentForms.xml"/>
+                                        <include-grid name="ListJobRequisitions" location="component://humanres/widget/forms/RecruitmentForms.xml"/>
                                     </decorator-section>
                                 </decorator-screen>
                             </widgets>
@@ -124,7 +124,7 @@ under the License.
                                         <include-form name="FindInternalJobPosting" location="component://humanres/widget/forms/RecruitmentForms.xml"/>
                                     </decorator-section>
                                     <decorator-section name="search-results">
-                                        <include-form name="ListInternalJobPosting" location="component://humanres/widget/forms/RecruitmentForms.xml"/>
+                                        <include-grid name="ListInternalJobPosting" location="component://humanres/widget/forms/RecruitmentForms.xml"/>
                                     </decorator-section>
                                 </decorator-screen>
                             </widgets>
@@ -187,7 +187,7 @@ under the License.
                                         <include-form name="FindJobInterview" location="component://humanres/widget/forms/RecruitmentForms.xml"/>
                                     </decorator-section>
                                     <decorator-section name="search-results">
-                                        <include-form name="ListInterview" location="component://humanres/widget/forms/RecruitmentForms.xml"/>
+                                        <include-grid name="ListInterview" location="component://humanres/widget/forms/RecruitmentForms.xml"/>
                                     </decorator-section>
                                 </decorator-screen>
                             </widgets>
@@ -249,7 +249,7 @@ under the License.
                                         <include-form name="FindApprovals" location="component://humanres/widget/forms/RecruitmentForms.xml"/>
                                     </decorator-section>
                                     <decorator-section name="search-results">
-                                        <include-form name="ListApprovals" location="component://humanres/widget/forms/RecruitmentForms.xml"/>
+                                        <include-grid name="ListApprovals" location="component://humanres/widget/forms/RecruitmentForms.xml"/>
                                     </decorator-section>
                                 </decorator-screen>
                             </widgets>
@@ -302,7 +302,7 @@ under the License.
                                         <include-form name="FindRelocation" location="component://humanres/widget/forms/RecruitmentForms.xml"/>
                                     </decorator-section>
                                     <decorator-section name="search-results">
-                                        <include-form name="ListRelocation" location="component://humanres/widget/forms/RecruitmentForms.xml"/>
+                                        <include-grid name="ListRelocation" location="component://humanres/widget/forms/RecruitmentForms.xml"/>
                                     </decorator-section>
                                 </decorator-screen>
                             </widgets>
--- a/applications/humanres/widget/forms/RecruitmentForms.xml
+++ b/applications/humanres/widget/forms/RecruitmentForms.xml
@@ -425,14 +425,14 @@ under the License.
                 <field-map field-name="viewSize" from-field="viewSize"/>
             </service>
        </actions>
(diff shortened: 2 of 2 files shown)


applications/humanres/widget/RecruitmentScreens.xml
```diff
@@ -249,7 +249,7 @@ under the License.
                                         <include-form name="FindApprovals" location="component://humanres/widget/forms/RecruitmentForms.xml"/>
                                     </decorator-section>
                                     <decorator-section name="search-results">
-                                        <include-form name="ListApprovals" location="component://humanres/widget/forms/RecruitmentForms.xml"/>
+                                        <include-grid name="ListApprovals" location="component://humanres/widget/forms/RecruitmentForms.xml"/>
```
danwatford, line 252: When testing with data from .`/gradlew loadAll`, plus some job requisitions created while testing earlier screens, I get the following error when searching from https://localhost:8443/humanres/control/FindApprovals.

Note: This error is not related to your PR, @PierreSmits , just mentioning it in case you knew the cause.


```
2022-01-21 15:38:55,042 |sse-nio-8443-exec-16 |ControlServlet                |E| Error in request handler:
java.lang.RuntimeException: Error rendering included form named [ListApprovals] at location [component://humanres/widget/forms/RecruitmentForms.xml]: java.lang.NullPointerExceptionjava.lang.NullPointerException
```

applications/humanres/widget/RecruitmentScreens.xml
```diff
@@ -302,7 +302,7 @@ under the License.
                                         <include-form name="FindRelocation" location="component://humanres/widget/forms/RecruitmentForms.xml"/>
                                     </decorator-section>
                                     <decorator-section name="search-results">
-                                        <include-form name="ListRelocation" location="component://humanres/widget/forms/RecruitmentForms.xml"/>
+                                        <include-grid name="ListRelocation" location="component://humanres/widget/forms/RecruitmentForms.xml"/>
```
danwatford, line 305: With this PR applied, the column header for search results on screen https://localhost:8443/humanres/control/FindRelocation is displayed as `${UILABELMAP.COMMONPOSITION`.

On trunk it is displayed as `EMPLOYEE NAME`.

PierreSmits:
Hi @danwatford<

Thanks for the comment. You mentioned that you created some test data to be used with this. Could you share that with the community, so that we can make an even better case why one should use OFBiz?

This may be the root cause of that error:
`      <field name="approverPartyId" field-name="partyId">
            <display-entity entity-name="PartyNameView" description="${firstName} ${lastName}">
                <sub-hyperlink target="/partymgr/control/viewprofile" target-type="inter-app" description="${approverPartyId}" link-style="buttontext">
                    <parameter param-name="partyId" from-field="approverPartyId"/>
                </sub-hyperlink>
            </display-entity>
        </field>`
Which can be found around line 366 in https://github.com/apache/ofbiz-framework/blob/trunk/applications/humanres/widget/forms/RecruitmentForms.xml

danwatford:
I just created a few dummy records as I went - but I can add something to the demo data, assuming that I properly understand the intent of the fields. Leave it with me.

applications/humanres/widget/forms/RecruitmentForms.xml
```diff
@@ -427,14 +425,14 @@ under the License.
...
-        <field name="partyId" title="Employee Name">
+        <field name="partyId" title="${UiLabelMap.CommonParty}">
             <display-entity entity-name="PartyNameView" description="${firstName} ${middleName} ${lastName} ${groupName}">
                 <sub-hyperlink target="/partymgr/control/viewprofile" target-type="inter-app" description="${partyId}" link-style="buttontext">              
                     <parameter param-name="partyId" from-field="partyId"/>
                 </sub-hyperlink>
             </display-entity>
         </field>
-        <field name="emplPositionId" title="${uiLabelMap.HumanResEmployeePositionId}" widget-style="linktext">
+        <field name="emplPositionId" title="${uiLabelMap.CommonPosition" widget-style="linktext">
```
danwatford: Missing closing brace following CommonPosition.

danwatford (commented):
Minor typo needs fixing, but otherwise checked and looks good.

PierreSmits:
My apologies for that, Daniel.

I have put in an amendment to correct that.

PierreSmits:
Hi @danwatford ,

As you can see from line item 429 the display entity for the field also show ${groupName}, meaning that also party groups may be referenced. Thus having CommonParty is a valid choice for the title. But more importantly the existing title of the field is **hard-coded**, not an existing label. So that is wrong. 
Additionally when data is available in the appropriate records, the value shown is a combination of **name** and **id** of the party. 
Hence also warranting an improvement. We also have to keep in mind that it is totally acceptable that external people through subcontracting can be in an employee position and a falling in a reporting structure.

As to change on line item 435, as a value a positionId is shown, but it is by implication a position.  A generic label for position (as it also can be used with lat and long, locations in warehouses etc) would make sense.

That Michael believes that hard coded titles or even field specific labels are good enough for the project, is his prerogative. He may think and believe so. 
But that does **NOT** imply that he can dictate to other contributors to throw out common sense and to do as he tells (as a suggestion, of course) and not apply that common sense to make OFBiz a better product.

The application of a common label has been around since day one of the project, and has been applied across all applications to be able to reduce the load on system, contributor, developer and network. If fields are used in forms across application then a common label is warranted to be preferred over a component specific, eg CommonProduct in grids and lookup fields, instead of productId.

Do a search/find in the labels for <property*productId", and you'll see I have a point (for improvement). And there are plenty more of such examples.

JacquesLeRoux:
> <field name="partyId" title="Employee Name" 

That's indeed wrong in any cases (in RecruitmentForms.xml), and dates from Jira OFBIZ-2205

danwatford:
Hi Pierre,

So my difficulty is assessing this PR is that it is changing two different levels of implementation.

The first category of changes are the switch from list forms to grids. This is a technical change where the same pattern of changes can be applied again and again with low risk, allowing the committer to easily comprehend intent.

As soon as we consider the second category of label changes, i.e. the application level, we then need the committer to understand the intention of the application - in this case the HR application and how job requisitions, applications, interviews and relocations work. I personally do not have a good knowledge of that application and have to spend a long time trying to figure out what the original intention was behind the application, and whether the application is broken in some way or whether there is a misunderstanding on the part of the contributor. See rules 1 and 2 here - https://cwiki.apache.org/confluence/display/OFBIZ/OFBiz+Committers+Roles+and+Responsibilities

The application intent also applies to labels. In the case of position as pertaining to employment, this is similar to the meaning of role. But in English position can also refer to location, status, etc. I only speak one language so have to err on the side of caution when generalising labels as I don't know if it will cause trouble for other languages.

In this specific case I thought it better to return to the hard-coded title, Employee Name, since it seemed the safest way of not breaking the intention of the HR application - or at least not making the intention any worse.

You mentioned the use of ${groupName} being an indicator that groups, in addition to individuals, can be provided relocation. I wouldn't have expected that to be the intention of the HR application in this case, so it may be the case that use of ${groupName} is incorrect in this case, rather than the use of the title 'Employee Name', accepting that hard-coded values are not great. I don't know which is the case and would need to investigate further.

Getting up to speed on the HR application will take time, therefore moving and discussing application intent changes to a different jira ticket or the mailing list would be my preference, hence my reverting of the labelling changes on this PR where there is the potential for ambiguity across languages.

mbrohl:
Thanks Daniel,
for your thorough description of the problem with common labels, I appreciate your time explaining it!

mbrohl:
@PierreSmits please stop putting words in my mouth which I never said, e.g. "hard coded titles.. are good enough for the project". They aren't.
But trying to fix a bad state with another bad solution is also not good enough for the project.

Introducing common or unspecific labels all over the place without context is simply wrong, as several other contributors pointed out in the past and recent code reviews and Jiras.

They are wrong in the aspect of differences of meaning between languages where context often matters and they also cut functionality by removing the possibility to change labels throughout applications and contexts. They making it harder for users to adjust OFBiz to their individual terminology which can be even different between divisions in the same company.

Those changes are not acceptable for the project. This is not only my personal viewpoint or "dictate" but the view of several others and I have not seen a single other contributor who is in favour of replacing specific labels with common labels in this broad approach.

PR #471, opened 2022-01-22
Improved: Agreement Roles - VIEW permissions (OFBIZ-12518)

PierreSmits:
Currently, a user with only 'VIEW' permissions, as demonstrated in trunk demo with userId = auditor, accessing the Agreement Roles screen, sees editable fields and/or triggers (to requests) reserved for users with 'CREATE' or 'UPDATE' permissions.

To see/test: https://localhost:8443/accounting/control/EditAgreementRoles?agreementId=8000

modified:
- AgreementScreens.xml - restructured screen EditAgreementRoles to work with permissions
- AgreementForms.xml - added grid AgreementRoles for users with VIEW permissions
additional cleanup


mbrohl:
Hi Pierre,

thanks for your contribution!

Why did you change the specific field translation labels to common labels, e.g. uiLabelMap.PartyPartyId vs uiLabelMap.CommonParty? The field contains the partyId which makes uiLabelMap.PartyPartyId the correct label.
The same applies to PartyRoleTypeId.

Can you please change this back to the specific label, thanks.

JacquesLeRoux:
Hi,

I agree with Michael, but we can do otherwise, reopening

PierreSmits:
Hi @JacquesLeRoux 

With your label corrections you neglected the aspect that in several other screens (and templates) CommonRole and CommonParty as labels already have been applied in various forms in the codebase. 

Are you with your corrections to this pull requests suggesting that contributors should favor archaic dogma over UX improvements?

PierreSmits (commented):
Hi @JacquesLeRoux,

CommonRole is already applied in following forms in the codebase:

- BillingAccountForms.xml
- FinAccoountForms.xml
- ProductionRunForms.xml
- RoutingTaskForms.xml
- OrderMenus.xml
- SubscriptionForms.xml
- ProjectForms.xml

CommonParty has been already applied in following forms and templates  in the codebase:

- FinAccountTrans.ftl
- BillingAccountForms.xml
- FinAccountForms.xml
- SubscriptionForms.xml

Anywhere where the user is seeing something more than just the primary Id field of a record, the label should reflect that.

Please revert your label changing commit in favour of enhancing the User Experience.

JacquesLeRoux:
@PierreSmits,

The reason I agreed with Michael in substance is because now OFBiz is really mature. So if you change a label id you also change the possible translation people have used on their side. I hope it clarifies my position

It was a time that we discussed about labels and their content. Paradoxally in 2009, about partyPartyId label we (French people) decided to replace all "Réf. acteur" by "Acteur". Since then we have also automatically added the names of the party. So it's OK.

I just found that we kept one for MarketingTrackingCodeVisitPartyId because there it makes sense. I could not find the discussion about that but https://issues.apache.org/jira/browse/OFBIZ-2958 is relevant.

PR #479, opened 2022-01-23
Improved: Have a Menu in SFA featuring actions to create the main objects (OFBIZ-12527)

PierreSmits:
Currently the create buttons for the main objects of the SFA application are located within find and profile widgets/templates of those objects.
In order to improve the usability of this application, OFBiz and thus the appeal of it for adopters and users, these create buttons/links/etc. should be in a main action menu visible at all times when a user with CREATE permissions is working within the application

modified:
- SfaMenus.xml - Added menu MainActionMenu
- included MainActionMenu in various screens to ensure that it is visible to user with CREATE permissions
- additional cleanup


PierreSmits:
Due to conflicts

PR #482, opened 2022-01-24
Improved: Accounting Rates - VIEW permissions (OFBIZ-12530)

PierreSmits:
Currently, a user with only 'VIEW' permissions, as demonstrated in trunk demo with userId = auditor, accessing the Rates screen, sees editable fields and/or triggers (to requests) reserved for users with 'CREATE' or 'UPDATE' permissions.

To see/test: https://localhost:8443/accounting/control/viewRateAmounts

Modified:
- GlobalGlAccountsScreens.xml - restructured screen ViewRateAmounts to work with permissions
- GlobalGlAccountsForms.xml - added grid RateAmounts for users with VIEW permissions
additional cleanup


mbrohl:
I think the removal of the blank lines between the elements makes the files less readable and maintanable.
It would also be better to split between functional changes and code cleanup, would make it possible to only take the functional change and leave the (maybe unwanted) "cleanups".

PierreSmits:
Thanks you for your opinion.

`It would also be better to split between functional changes and code cleanup, would make it possible to only take the functional change and leave the (maybe unwanted) "cleanups".`
If you feel that strongly about it, go write a proposal and send that to the dev mailing list and start a discussion there.

PR #483, opened 2022-01-25
Improved: Accounting Transactions - VIEW permissions (OFBIZ-12532)

PierreSmits:
Currently, a user with only 'VIEW' permissions, as demonstrated in trunk demo with userId = auditor, accessing the accounting transactions screen, sees editable fields and/or triggers (to requests) reserved for users with 'CREATE' or 'UPDATE' permissions in unposted transactions.

To see/test with an unposed transaction: https://localhost:8443/accounting/control/EditAcctgTrans?organizationPartyId=Company&acctgTransId=10000

modified: GlScreens.xml
- restructured screen EditAcctgTrans to work with permissions
- additional cleanup and enhancements vis-a-vis consistency.

### DIFF
--- a/applications/accounting/widget/GlScreens.xml
+++ b/applications/accounting/widget/GlScreens.xml
@@ -20,10 +20,14 @@ under the License.
 
 <screens xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xmlns="http://ofbiz.apache.org/Widget-Screen" xsi:schemaLocation="http://ofbiz.apache.org/Widget-Screen http://ofbiz.apache.org/dtds/widget-screen.xsd">
-
     <screen name="CommonPartyGlDecorator">
         <section>
             <actions>
+                <property-map resource="CommonUiLabels" map-name="uiLabelMap" global="true"/>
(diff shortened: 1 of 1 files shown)


mbrohl:
I think the removal of the blank lines between the elements makes the files less readable and maintanable.
It would also be better to split between functional changes and code cleanup, would make it possible to only take the functional change and leave the (maybe unwanted) "cleanups".

JacquesLeRoux:
Hi Michael,

>I think the removal of the blank lines between the elements makes the files less readable and maintanable.

I initially had the same thought, because I think it's indeed less readable, add nothing and rather complicate reviews. But I know there are parts where it's done the way or the other. So I let it be. I'd not advice to set a rule for that, else, **in the absolute**, we would have to change everywhere it's already not conform to the new rule.

This said we can ask Pierre to remove these kind of changes and don't complete reviews if we see things diverging.

mbrohl:
I did not intend to set a rule on this, I just gave my opinion as a reviewer.
I would appreciate if those changes would be removed, though. There's always hope...

JacquesLeRoux:
Hope is a fragile plant :smirk:

PR #486, opened 2022-01-25
Improved: CommonStoreDecorator - consistency (OFBIZ-12537)

PierreSmits:
The common decoratr screen for store records has the title and the action menu for users with CREATE and/or UPDATE permissions in places inconsistent to similar screens for other objects. The elements should move to improve the consistency of the user experience.

modified: CommonScreens.xml
relocated title and the object action menu for a store record.
additional cleanup


JacquesLeRoux:
After comparing with demo, I prefer the current one and see no advantages with that, closing

PR #494, opened 2022-01-27
Fixed: Wrong Field Name Definition in RequirementForms (OFBIZ-12505)

PierreSmits:
The issue exists also in the older version.
The field "requirementByDate" form "FindApprovedProductRequirements" in RequirementForms.xml has been set wrongly and it should be "requiredByDate".
This causes the result could not be queried out properly as indicated as below:
The result is governed by the service "getRequirementsForSupplier", where reads the list from the view "RequirementAndRole".
In the view entity definition, the field is defined as
`<alias entity-alias="RQ" name="requiredByDate"/>, but it is not the case in the form "FindApprovedProductRequirements" `above.

Modified: RequirementForms.xml
- changed field requirementByDate into requiredByDate as per entity definition for Requirement

Thanks to Schumann Ye for reporting

### DIFF
--- a/applications/order/widget/ordermgr/RequirementForms.xml
+++ b/applications/order/widget/ordermgr/RequirementForms.xml
@@ -282,7 +282,7 @@ under the License.
             </drop-down>
         </field>
         <field name="productId" title="${uiLabelMap.ProductProductId}"><lookup target-form-name="LookupProduct"/></field>
-        <field name="requirementByDate"><date-find type="date"/></field>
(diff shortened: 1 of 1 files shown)


mbrohl:
Please, Pierre, use git pull --rebase to merge trunk in your feature branches to avoid those nasty empty merge commits. Thanks!

PierreSmits:
Thanks Michael,

Will do. I guess you meant perform a **git pull asf trunk --rebase**

JacquesLeRoux:
Fixed directly in OFBIZ-12505

PR #498, opened 2022-02-03
Improved: WorkEffort - MainActionMenu (OFBIZ-12557)

PierreSmits:
Currently the create buttons for the main objects of the work effort  are located within the find and other widgets/templates of those objects.
In order to improve the usability of OFBiz (and thus the appeal of it for adopters and users) these create buttons/links/etc. should be in a main action menu visible at all times when a user is working within the component.

modified:
WorkEffortMenus.xml - added MainActionMenu for users with CREATE permission in the component
CommonScreens.xml - added MainActionMenu as an 'include-menu' ref in various common decorator screens.
Added MainActionMenu as an 'include-menu' in screen definitions in various other Screen.xml files where appropriate
additional cleanup


mbrohl:
As said before, the removal of blank lines between sections makes the files harder to read.

PR #509, opened 2022-02-09
Improved: Biling Account Payment - VIEW Permissions (OFBIZ-12565)

PierreSmits:
When accessing
[https://localhost:8443/accounting/control/BillingAccountPayments?billingAccountId=9010]
as a user with only VIEW permissions (e.g. userid=auditor),
the screen shows a form to create a payment.
This should not be visible to such a user as it leads
to a undesired effect and diminished user experience.

modified: BillingAccountScreens.xml
reworked screen BillingAccountPayments to work with
CREATE permissions


JacquesLeRoux:
The Jira issue number was removed from title: wrong

PR #513, opened 2022-04-13
Improved: In UtilHttp, for regex processing of urls, replace Java regex with RE2J (OFBIZ-12599)

jacopoc:
The Main advantage of RE2J is that it guarantees linear time execution.

Note that it has not a complete API: https://github.com/google/re2j#caveats


mbrohl:
Hi Jacopo,

can you explain a bit more what this change offers and why it warrants the introduction of a new dependency?
Do we have a perfomance issue?

PR #515, opened 2022-06-15
Improved : The typing errors in the file (OFBIZ-12643)

priyal-hotwax:
Fixed the typing errors in the file.

Thanks: Priyal Jain, Priya Sharma and Aditya

### DIFF
--- a/framework/webapp/dtd/site-conf.xsd
+++ b/framework/webapp/dtd/site-conf.xsd
@@ -477,7 +477,7 @@ under the License.
         <xs:attribute name="name" type="xs:string" use="required">
             <xs:annotation>
                 <xs:documentation>
-                    The name of the response, which also matches the string returned by the event
+                    The name of the response, that also matches the string returned by the event
                 </xs:documentation>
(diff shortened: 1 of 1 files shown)


priyasharma1 (approved):

JacquesLeRoux:
Hi Priya,

While at ut, I'd rather use "that" than "which", ie
> The name of the response, which also matches the string returned by the event

>The name of the response, that also matches the string returned by the event

And the commit title and comment should be conform

TIA

framework/webapp/dtd/site-conf.xsd
```diff
@@ -477,7 +477,7 @@ under the License.
         <xs:attribute name="name" type="xs:string" use="required">
             <xs:annotation>
                 <xs:documentation>
-                    The name of the response, wich also matches the string returned by the event
+                    The name of the response, which also matches the string returned by the event
```
adityasharma7:
```suggestion
                    The name of the response, that also matches the string returned by the event
```

PR #531, opened 2022-08-01
Improved: findParty_Status. (OFBIZ-12677)

georg1312:
Improved: Backend Party Manager: Removed hidden statusId field and added drop-down for status selection in findparty form to be able to find not ENABLED parties too. (OFBIZ-12677)

Central actor search around the static Party.statusId field simplified and the logic behind it fixed. Thus, the statusId value is also available in performPartyFind Service.

### DIFF
--- a/applications/party/servicedef/services_view.xml
+++ b/applications/party/servicedef/services_view.xml
@@ -74,7 +74,6 @@ under the License.
     <service name="performFindParty" engine="java"
         location="org.apache.ofbiz.party.party.PartyServices" invoke="performFindParty">
         <description>General Party Find Service, duplicated for screen widget purpose, Used in the new findparty page in the Party Manager, etc</description>
-        <attribute name="roleTypeId" type="String" mode="IN" optional="true"/> <!-- can be null or ANY to include any -->
         <attribute name="noConditionFind" type="String" mode="IN" optional="true" default-value="N"/>
         <attribute name="extInfo" type="String" mode="IN" optional="true"/>
         <attribute name="extCond" type="org.apache.ofbiz.entity.condition.EntityCondition" mode="IN" optional="true">
@@ -82,6 +81,8 @@ under the License.
         </attribute>
         <attribute name="partyId" type="String" mode="IN" optional="true"/> <!-- does a LIKE compare on this, can do partial, case insensitive, etc -->
         <attribute name="partyTypeId" type="String" mode="IN" optional="true"/>
+        <attribute name="roleTypeId" type="String" mode="IN" optional="true"/>
+        <attribute name="statusId" type="String" mode="IN" optional="true"/>
         <attribute name="userLoginId" type="String" mode="IN" optional="true"/> <!-- does a LIKE compare on this, can do partial, case insensitive, etc -->
         <attribute name="externalId" type="String" mode="IN" optional="true"/>
         <attribute name="groupName" type="String" mode="IN" optional="true"/>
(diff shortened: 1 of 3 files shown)


applications/party/servicedef/services_view.xml
```diff
@@ -74,14 +74,15 @@ under the License.
...
         <description>General Party Find Service, duplicated for screen widget purpose, Used in the new findparty page in the Party Manager, etc</description>
-        <attribute name="roleTypeId" type="String" mode="IN" optional="true"/> <!-- can be null or ANY to include any -->
         <attribute name="noConditionFind" type="String" mode="IN" optional="true" default-value="N"/>
         <attribute name="extInfo" type="String" mode="IN" optional="true"/>
         <attribute name="extCond" type="org.apache.ofbiz.entity.condition.EntityCondition" mode="IN" optional="true">
             <description>EntityCondition that can be send to this service to manage complex search case</description>
         </attribute>
         <attribute name="partyId" type="String" mode="IN" optional="true"/> <!-- does a LIKE compare on this, can do partial, case insensitive, etc -->
         <attribute name="partyTypeId" type="String" mode="IN" optional="true"/>
+        <attribute name="roleTypeId" type="String" mode="IN" optional="true"/>
```
adityasharma7, line 84: I think this is unwanted change. Could you please revert?

applications/party/src/main/java/org/apache/ofbiz/party/party/PartyServices.java
```diff
@@ -1604,14 +1604,17 @@ public static Map<String, Object> performFindParty(DispatchContext dctx, Map<Str
...
 
-        // now the statusId - send ANY for all statuses; leave null for just enabled; or pass a specific status
+        // now the statusId - null for all statuses; "PARTY_ENABLED" finds statusId == null entries also, other passed in values are matched exactly
         if (UtilValidate.isNotEmpty(statusId)) {
-            andExprs.add(EntityCondition.makeCondition("statusId", statusId));
-        } else {
-            // NOTE: _must_ explicitly allow null as it is not included in a not equal in many databases... odd but true
-            andExprs.add(EntityCondition.makeCondition(EntityCondition.makeCondition("statusId", GenericEntity.NULL_FIELD),
-                    EntityOperator.OR, EntityCondition.makeCondition("statusId", EntityOperator.NOT_EQUAL, "PARTY_DISABLED")));
+            if ("PARTY_ENABLED".equals(statusId)) {
```
adityasharma7, line 1609: Could you please share use case why this change is needed?

adityasharma7 (changes requested):

applications/party/servicedef/services_view.xml
JacquesLeRoux: I agree I see no reason why
        `<attribute name="roleTypeId" type="String" mode="IN" optional="true"/> <!-- can be null or ANY to include `any -->
has been moved. Not a big deal but we lose the comment

applications/party/src/main/java/org/apache/ofbiz/party/party/PartyServices.java
JacquesLeRoux: I think it's OK to be able to select disabled party. That can make sense in some situations.

mbrohl: We'll discuss those changes and will provide an altered pull request. 
Will take at least two weeks as georg is not available atm.

JacquesLeRoux:
Hi @mbrohl, should we close?

mbrohl:
I'll check when I'm back from vacation.

PR #541, opened 2022-08-26
OFBIZ-12680 Optimized the 'get average product rating' method logic.

sourabhjains04:
Done following:

- Enhanced and optimized the OOTB 'get average product rating' method logic and updated its references.
- Updated the 'get average product rating' method logic and optimized the logic by using a dynamic entity using the 'average' function to get the average rather than iterating all the product reviews.
- Updated the service parameter and updated all the references to add 'delegator' as the new parameter.

Improved: (OFBIZ-12680)
Explanation: Optimized the 'get average product rating' method logic.


JacquesLeRoux:
Closing, this is done at OFBIZ-12680

PR #547, opened 2022-09-29
Improved: Support title attribute on hyperlinks (OFBIZ-12697)

florianMo:
Improved: Support title attribute on hyperlink, using FlexibleStringExpander

(OFBIZ-12697)

Explanation : title attribute exists in XSD for hyperlinks, but it is not used as a title attribute value on the rendered anchor tag (<a title="..."/>). With this PR, title attribute of hyperlinks is used as a title attribute on <a/>, so we can control the native browser tooltip which appears when hovering the link. Works for anchor-link and form submission hyperlinks.

Thanks: Néréide Team :)

### DIFF
--- a/framework/widget/src/main/java/org/apache/ofbiz/widget/model/CommonWidgetModels.java
+++ b/framework/widget/src/main/java/org/apache/ofbiz/widget/model/CommonWidgetModels.java
@@ -330,6 +330,7 @@ public final class CommonWidgetModels {
         private final boolean encode;
         private final boolean fullPath;
         private final FlexibleStringExpander idExdr;
+        private final String title;
         private final Image image;
         private final String linkType; // anchor or hidden form
         private final FlexibleStringExpander nameExdr;
@@ -352,6 +353,7 @@ public final class CommonWidgetModels {
         public Link(Element linkElement) {
             this.textExdr = FlexibleStringExpander.getInstance(linkElement.getAttribute("text"));
             this.idExdr = FlexibleStringExpander.getInstance(linkElement.getAttribute("id"));
+            this.title = linkElement.getAttribute("title");
             this.styleExdr = FlexibleStringExpander.getInstance(linkElement.getAttribute("style"));
             this.nameExdr = FlexibleStringExpander.getInstance(linkElement.getAttribute("name"));
             this.targetExdr = FlexibleStringExpander.getInstance(linkElement.getAttribute("target"));
@@ -426,6 +428,7 @@ public final class CommonWidgetModels {
             this.encode = false;
             this.fullPath = false;
             this.idExdr = FlexibleStringExpander.getInstance("");
+            this.title = "";
             this.image = null;
             this.linkType = "";
             this.nameExdr = FlexibleStringExpander.getInstance("");
@@ -452,6 +455,7 @@ public final class CommonWidgetModels {
             this.encode = false;
             this.fullPath = false;
             this.idExdr = FlexibleStringExpander.getInstance("");
+            this.title = "";
             this.image = null;
             this.linkType = "";
             this.nameExdr = FlexibleStringExpander.getInstance("");
@@ -508,6 +512,10 @@ public final class CommonWidgetModels {
 
         public FlexibleStringExpander getIdExdr() {
             return idExdr;
+        }
+
+        public String getTitle() {
+            return this.title;
         }
 
         public Image getImage() {
--- a/framework/widget/src/main/java/org/apache/ofbiz/widget/renderer/macro/MacroFormRenderer.java
+++ b/framework/widget/src/main/java/org/apache/ofbiz/widget/renderer/macro/MacroFormRenderer.java
@@ -215,6 +215,7 @@ public final class MacroFormRenderer implements FormStringRenderer {
         this.request.setAttribute("imageTitle", encodedImageTitle);
         this.request.setAttribute("descriptionSize", hyperlinkField.getSize());
         this.request.setAttribute("id", modelFormField.getCurrentContainerId(context));
+        this.request.setAttribute("title", hyperlinkField.getTitle());
         this.request.setAttribute("width", hyperlinkField.getWidth());
         this.request.setAttribute("height", hyperlinkField.getHeight());
         makeHyperlinkByType(writer, hyperlinkField.getLinkType(), modelFormField.getWidgetStyle(), hyperlinkField.getUrlMode(),
@@ -2971,12 +2972,13 @@ public final class MacroFormRenderer implements FormStringRenderer {
             String event = "";
             String action = "";
             String imgSrc = "";
+            String imgTitle = "";
             String alt = "";
             String id = "";
             String uniqueItemName = "";
             String width = "";
             String height = "";
-            String imgTitle = "";
+            String title = "";
             String hiddenFormName = WidgetWorker.makeLinkHiddenFormName(context, modelFormField);
             if (UtilValidate.isNotEmpty(modelFormField.getEvent()) && UtilValidate.isNotEmpty(modelFormField.getAction(context))) {
                 event = modelFormField.getEvent();
@@ -2985,22 +2987,22 @@ public final class MacroFormRenderer implements FormStringRenderer {
             if (UtilValidate.isNotEmpty(request.getAttribute("image"))) {
                 imgSrc = request.getAttribute("image").toString();
             }
-            if (UtilValidate.isNotEmpty(request.getAttribute("alternate"))) {
-                alt = request.getAttribute("alternate").toString();
-            }
             if (UtilValidate.isNotEmpty(request.getAttribute("imageTitle"))) {
                 imgTitle = request.getAttribute("imageTitle").toString();
             }
-            Integer size = Integer.valueOf("0");
+            if (UtilValidate.isNotEmpty(request.getAttribute("alternate"))) {
+                alt = request.getAttribute("alternate").toString();
+            }
+            int size = 0;
             if (UtilValidate.isNotEmpty(request.getAttribute("descriptionSize"))) {
-                size = Integer.valueOf(request.getAttribute("descriptionSize").toString());
+                size = Integer.parseInt(request.getAttribute("descriptionSize").toString());
             }
+            // if description is truncated, always use description as title
             if (UtilValidate.isNotEmpty(description) && size > 0 && description.length() > size) {
-                imgTitle = description;
-                description = description.substring(0, size - 8) + "..." + description.substring(description.length() - 5);
-            }
-            if (UtilValidate.isEmpty(imgTitle)) {
-                imgTitle = modelFormField.getTitle(context);
+                title = description;
+                description = description.substring(0, size) + "…";
+            } else if (UtilValidate.isNotEmpty(request.getAttribute("title"))) {
+                title = request.getAttribute("title").toString();
             }
             if (UtilValidate.isNotEmpty(request.getAttribute("id"))) {
                 id = request.getAttribute("id").toString();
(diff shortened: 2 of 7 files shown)


framework/widget/src/main/java/org/apache/ofbiz/widget/model/CommonWidgetModels.java
```diff
@@ -510,6 +514,14 @@ public FlexibleStringExpander getIdExdr() {
             return idExdr;
         }
 
+        public String getTitle(Map<String, Object> context) {
+            return this.titleExdr.expandString(context);
+        }
+
+        public FlexibleStringExpander getTitleExdr() {
```
danwatford: I don't think this is needed. It is good to reduce the size of the public interface where possible. 

We can always add it back in later if a requirement materialises, but later removing a public interface can be much more problematic.

framework/widget/src/main/java/org/apache/ofbiz/widget/renderer/macro/MacroFormRenderer.java
```diff
@@ -3005,6 +3006,9 @@ public void makeHyperlinkString(Appendable writer, String linkStyle, String targ
             if (UtilValidate.isNotEmpty(request.getAttribute("id"))) {
                 id = request.getAttribute("id").toString();
             }
+            if (UtilValidate.isNotEmpty(request.getAttribute("title"))) {
+                imgTitle = request.getAttribute("title").toString();
```
danwatford: Given the number of conditional assignments to the imgTitle variable in this method, have you been able to infer any precedence rules for how the title attribute should be determined? 

If so it would be ideal to add tests to demonstrate those rules in MacroFormRendererTest. This would help with the ongoing refactoring effort of MacroFormRenderer to RenderableFtlFormElementsBuilder.

danwatford: We should also take this opportunity to rename the variable from `imgTitle` to `title`, better indicating the intent of the variable to the reader.

danwatford:
Hyperlink rendering in OFBiz is not the most friendly code to work with. I've had my head spinning many times when dealing with it, and haven't got close to any refactoring on that aspect of form rendering yet.

Kudos for giving it a go, @florianMo  :)

danwatford (commented):
@florianMo please could you also update my `hyperlinkMacroCall` in RenderableFtlFormElementsBuilder as some of the hyperlink rendering code has been replicated there as part of refactoring MacroFormRenderer (OFBIZ-11456)

framework/widget/src/main/java/org/apache/ofbiz/widget/renderer/macro/MacroFormRenderer.java
florianMo: @danwatford I came up with this logic : 

- do not mix `image-title` (which can only be rendered as the img title) and `title` (which can only be rendered as the link title)
- if the link's text content is truncated (`description` longer than the `size` attribute), then use the full `description` as link title (we need to give the user the opportunity to access to the the full text content)
- otherwise, use the provided `title` as link's title

I think it's easier to understand this way, what do you think ?

florianMo:
> @florianMo please could you also update my `hyperlinkMacroCall` in RenderableFtlFormElementsBuilder as some of the hyperlink rendering code has been replicated there as part of refactoring MacroFormRenderer (OFBIZ-11456)

@danwatford I replicated the modifications I made in `MacroFromRenderer.java` into `RenderableFtlFormElementsBuilder.java`. Let me know if I misunderstood (still discovering OFBiz :).

framework/widget/src/test/java/org/apache/ofbiz/widget/renderer/macro/MacroFormRendererTest.java
```diff
@@ -887,6 +887,55 @@ public void renderSortFieldUsesQueryString(@Mocked ModelForm modelForm) throws I
...
+        new Expectations() {
+            {
+                hyperlinkField.getDescription(withNotNull()); result = description;
+                hyperlinkField.getTarget(withNotNull()); result = "#";
+                request.getAttribute("descriptionSize"); result = 5;
+            }
+        };
+
+        macroFormRenderer.renderHyperlinkField(appendable, new HashMap<>(), hyperlinkField);
+        assertAndGetMacroString("makeHyperlinkString", ImmutableMap.of("description", "DESCR…", "title", description));
```
danwatford, line 938: This test's name suggests proving that '#' is rendered for the title attribute, but instead it is proving that 'DESCRIPTION' is rendered. Was that the intention?

framework/widget/src/main/java/org/apache/ofbiz/widget/renderer/macro/MacroFormRenderer.java
danwatford: The logic sounds good, and even better is that you have asserted it with tests.
Sorry to add to your work, but could you see if you can create the corresponding tests in RenderableFtlFormElementsBuilderTest.

danwatford (commented):
Implementation looks good to me. Just a couple of comments left regarding tests.

framework/widget/src/test/java/org/apache/ofbiz/widget/renderer/macro/MacroFormRendererTest.java
florianMo: The `#` is provided for `target` attribute, to avoid a NPE in the rendering process (`buildHyperlinkUri` calls `Parser.unescapeEntities` on `target`).

But indeed something is unclear here. If description is truncated, then request.getAttribute("title") is not even called. I'll add `hyperlinkField.getTitle(); result = title;` to make clear that we provide a `title`, which is overriden by `description` because we have an ellipsis.

danwatford: Ah I confused target and title!

Anyway, test hyperlinkFieldMacroRenderedTruncatedWithTitle seems to have the same outcome as test hyperlinkFieldMacroRenderedTruncatedNoTitle. Was that the intention?

florianMo:
@danwatford reagrding your last comment (adding my tests in `RenderableFtlFormElementsBuilderTest`), it seems that it would take some rewriting, I would have expected that `makeHyperlinkString` would take a `ModelFormField.HyperLinkField`, but it takes a `ModelFormField.SubHyperlink` instead, so I'm not sure how I could test it. Is it an actual error?

danwatford:
> @danwatford reagrding your last comment (adding my tests in `RenderableFtlFormElementsBuilderTest`), it seems that it would take some rewriting, I would have expected that `makeHyperlinkString` would take a `ModelFormField.HyperLinkField`, but it takes a `ModelFormField.SubHyperlink` instead, so I'm not sure how I could test it. Is it an actual error?

Hi @florianMo ,

It could well be an error, but might also just be an oversight. I found the hyperlink code quite difficult to follow and hoped it would become easier to deal with once the more refactoring was complete with ModelFormRenderer.

The good thing is that your work prompted me to take another look at how links are represented so I feel I have a better handle on them now.  There are also a few deprecations, dated back at least to 2016, related to hyperlink attributes in the .xsd files so we might find removing those simplifies things.

**I withdraw my request that you add tests `RenderableFtlFormElementsBuilderTest`** as I am happy you have encoded rules around the title attribute in MacroFormRendererTest.

framework/widget/src/main/java/org/apache/ofbiz/widget/renderer/macro/MacroFormRenderer.java
danwatford: I'm no longer requesting that tests are added to RenderableFtlFormElementsBuilderTest.

florianMo:
ok @danwatford ! I saw that image related attributes were now deprecated, that's why I chose to isolate them from the link title question (`image-title` and `title` were mixed in the logic).

framework/widget/src/test/java/org/apache/ofbiz/widget/renderer/macro/MacroFormRendererTest.java
florianMo: - `hyperlinkFieldMacroRenderedTruncatedNoTitle` : we use full `description` as title if `description` is truncated (hovering "DESC…" will show "DESCRIPTION")
- `hyperlinkFieldMacroRenderedTruncatedWithTitle` : user provided a `title` attribute, but since the text content of the link (`description`) is truncated, we use untruncated `description` as the title attribute (hovering "DESC…" will show "DESCRIPTION", and the provided `title` won't be used). This is an opinionated choice, we can choose the other way around (always use `title` for title attribute) if it feels more logical ?

So indeed these 2 tests have the same outcome but for different test conditions.

danwatford:
Tested by observing how changes to form definitions affect displayed items.
Using default data.

Page: https://localhost:8443/partymgr/control/showclassgroups

The form on this page is defined by party/widget/partymgr/PartyClassificationForms.xml#ListPartyClassificationGroups

### Field: partyClassificationGroupId
Changes to the title, description and size attributes of the hyperlink element were all reflected as expected in the browser.
 
### Field: deleteLink
**Attribute `size` does not affect rendering of this hyperlink, meaning the description is not truncated.**

When not considering the `size` attribute, the `title` attribute works as expected.

#### Cause
The cause of the discrepancy between the two renderings of hyperlink fields was tracked down the deleteLink being rendered as a hidden form. When rendering these forms, the anchor tag is created in WidgetWorker's `makeHiddenFormLinkAnchorElement` method.

We therefore need the logic for rendering the anchor's title attribute based on the description, size and title attributes of the hyperlink field, similar to that used in MacroFormRenderer's `makeHyperlinkString` method.

florianMo:
> Tested by observing how changes to form definitions affect displayed items. Using default data.
> 
> Page: https://localhost:8443/partymgr/control/showclassgroups
> 
> The form on this page is defined by party/widget/partymgr/PartyClassificationForms.xml#ListPartyClassificationGroups
> ### Field: partyClassificationGroupId
> 
> Changes to the title, description and size attributes of the hyperlink element were all reflected as expected in the browser.
> ### Field: deleteLink
> 
> **Attribute `size` does not affect rendering of this hyperlink, meaning the description is not truncated.**
> 
> When not considering the `size` attribute, the `title` attribute works as expected.
> #### Cause
> 
> The cause of the discrepancy between the two renderings of hyperlink fields was tracked down the deleteLink being rendered as a hidden form. When rendering these forms, the anchor tag is created in WidgetWorker's `makeHiddenFormLinkAnchorElement` method.
> 
> We therefore need the logic for rendering the anchor's title attribute based on the description, size and title attributes of the hyperlink field, similar to that used in MacroFormRenderer's `makeHyperlinkString` method.

Good catch, checking it. Does OFBiz has some testing tools able to find these regressions, or is it just manual testing that led you there ?

florianMo:
@danwatford regarding the `size` attribute, I will fix this in #548 if it's ok with you (I prefer to keep the description/title/size business in the same PR).

florianMo:
@danwatford I fixed the hidden form link anchor case

danwatford:
> Good catch, checking it. Does OFBiz has some testing tools able to find these regressions, or is it just manual testing that led you there ?

I'm not aware of any tests that would address this particular area, but I haven't dug into the various integration tests in-depth.

If you search the sources for `--test` you will find a few examples in README.adoc which describes how to launch various integration tests (as opposed to unit tests).

The integration tests are surfaced through the various `ofbiz-component.xml` files which every component has. Components that support integration testing will have a `<test-suite />` element which describes the location of the tests. See framework/widget/ofbiz-component.xml for an example.

In this case I tested manually.

danwatford:
Testing with by observing changes to page https://localhost:8443/partymgr/control/showclassgroups when applying title, description and size attribute changes to hyperlink fields in party/widget/partymgr/PartyClassificationForms.xml#ListPartyClassificationGroups

### Field: partyClassificationGroupId
Changes to the title, description and size attributes of the hyperlink element were all reflected as expected in the browser.

### Field: deleteLink
Changes to the title, description and size attributes of the hyperlink element were all reflected as expected in the browser.

## Test passed.

PR #548, opened 2022-10-03
Improved: Use FlexibleStringExpander for disabled attribute on fields

florianMo:
Add disabled support on all form field types

Switch to `FlexibleStringExpander` (instead of boolean) to support dynamic attribute values

[OFBIZ-12678](https://issues.apache.org/jira/browse/OFBIZ-12678)

Improved: Add `disabled` support on all form field types 
Fixed: bug in `tab-index` attribute implementation for some HTML macros

### Explanation

Before this PR, `disabled` attribute on form fields is a `boolean`, and its support is not implemented on all eligible widgets : 
```
<text disabled="false" />
<text disabled="true" />
```

Based on https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled, this PR implements dynamic `disabled` attribute for all relevant widgets : 
```
<date-range disabled="true" /> // disabled
<date-range disabled="${5==5}" /> // disabled
<date-range disabled="false" /> // not disabled
<date-range disabled="${5==6}" /> // not disabled
```

In addition, in HTML macros, the `disabled` attribute is rendered as a `boolean` instead of a string (see https://www.w3.org/TR/1998/REC-DOM-Level-1-19981001/level-one-html.html#ID-642250288) : 
```
<text disabled="true" /> // renders as : 
<input disabled /> // instead of
<input disabled="true" />
```

Thanks: Néréide team

### DIFF
--- a/framework/widget/dtd/widget-form.xsd
+++ b/framework/widget/dtd/widget-form.xsd
@@ -857,7 +857,15 @@ under the License.
                     </xs:documentation>
                 </xs:annotation>
             </xs:attribute>
-            <xs:attribute name="disabled" type="xs:boolean" default="false"/>
+            <xs:attribute type="xs:string" name="disabled">
+                <xs:annotation>
+                    <xs:documentation>
+                        Set a field as disabled : not mutable, not focusable, not submitted if in a form.
+                        Accepts ${} notation to allow use of expressions.
+                        Expressions must evaluate to the strings 'true', 'false' (without quotes) or empty (equivalent to false)
+                    </xs:documentation>
+                </xs:annotation>
+            </xs:attribute>
         </xs:complexType>
     </xs:element>
 
--- a/framework/widget/src/main/java/org/apache/ofbiz/widget/model/ModelFormField.java
+++ b/framework/widget/src/main/java/org/apache/ofbiz/widget/model/ModelFormField.java
@@ -160,7 +160,7 @@ public final class ModelFormField {
     private final String parentFormName;
     private final String tabindex;
     private final String conditionGroup;
-    private final boolean disabled;
+    private final FlexibleStringExpander disabled;
 
     private ModelFormField(ModelFormFieldBuilder builder) {
         this.action = builder.getAction();
@@ -215,7 +215,7 @@ public final class ModelFormField {
         this.parentFormName = builder.getParentFormName();
         this.tabindex = builder.getTabindex();
         this.conditionGroup = builder.getConditionGroup();
-        this.disabled = builder.getDisabled();
+        this.disabled = builder.getDisabledSpec();
     }
 
     public FlexibleStringExpander getAction() {
@@ -483,8 +483,15 @@ public final class ModelFormField {
         return conditionGroup;
     }
 
-    public boolean getDisabled() {
+    public FlexibleStringExpander getDisabledSpec() {
         return disabled;
+    }
+
+    public boolean getDisabled(Map<String, Object> context) {
+        if (UtilValidate.isNotEmpty(this.disabled)) {
+            return "true".equals(disabled.expandString(context));
+        }
+        return false;
     }
 
     public Map<String, ? extends Object> getMap(Map<String, ? extends Object> context) {
@@ -2738,25 +2745,30 @@ public final class ModelFormField {
      */
     public static class HiddenField extends FieldInfo {
         private final FlexibleStringExpander value;
+        private final FlexibleStringExpander disabled;
 
         public HiddenField(Element element, ModelFormField modelFormField) {
             super(element, modelFormField);
             this.value = FlexibleStringExpander.getInstance(element.getAttribute("value"));
+            this.disabled = FlexibleStringExpander.getInstance(element.getAttribute("disabled"));
         }
 
         private HiddenField(HiddenField original, ModelFormField modelFormField) {
             super(original.getFieldSource(), original.getFieldType(), modelFormField);
             this.value = original.value;
+            this.disabled = original.disabled;
         }
 
         public HiddenField(int fieldSource, ModelFormField modelFormField) {
             super(fieldSource, FieldInfo.HIDDEN, modelFormField);
             this.value = FlexibleStringExpander.getInstance("");
+            this.disabled = FlexibleStringExpander.getInstance("");
         }
 
         public HiddenField(ModelFormField modelFormField) {
             super(FieldInfo.SOURCE_EXPLICIT, FieldInfo.HIDDEN, modelFormField);
             this.value = FlexibleStringExpander.getInstance("");
+            this.disabled = FlexibleStringExpander.getInstance("");
         }
 
         @Override
@@ -2797,6 +2809,18 @@ public final class ModelFormField {
                 return valueEnc;
             }
             return getModelFormField().getEntry(context);
+        }
+
+        /**
+         *
+         * @param context the context
+         * @return evaluated value
+         */
+        public boolean getDisabled(Map<String, Object> context) {
+            if (UtilValidate.isNotEmpty(this.disabled)) {
+                return "true".equals(disabled.expandString(context));
+            }
+            return false;
         }
 
         @Override
--- a/framework/widget/src/main/java/org/apache/ofbiz/widget/model/ModelFormFieldBuilder.java
+++ b/framework/widget/src/main/java/org/apache/ofbiz/widget/model/ModelFormFieldBuilder.java
@@ -116,7 +116,7 @@ public class ModelFormFieldBuilder {
     private String parentFormName = "";
     private String tabindex = "";
     private String conditionGroup = "";
-    private boolean disabled = false;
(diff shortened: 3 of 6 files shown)


framework/widget/src/main/java/org/apache/ofbiz/widget/model/ModelFormField.java
```diff
@@ -483,10 +483,17 @@ public String getConditionGroup() {
         return conditionGroup;
     }
 
-    public boolean getDisabled() {
+    public FlexibleStringExpander getDisabled() {
         return disabled;
     }
 
+    public String getDisabled(Map<String, Object> context) {
```
danwatford: Can we have a boolean returned here?   ModelFormField is the authority on whether a field is disabled based on context. If the property value is empty then ModelFormField can consider the field as enabled.

framework/widget/src/main/java/org/apache/ofbiz/widget/model/ModelFormField.java
```diff
@@ -483,10 +483,17 @@ public String getConditionGroup() {
         return conditionGroup;
     }
 
-    public boolean getDisabled() {
+    public FlexibleStringExpander getDisabled() {
```
danwatford: Perhaps rename this method to something like `getDisabledSpec` to indicate that it now contains the specification of how the enabled/disabled state of a field will be determined.

If possible, keep this field private as caller should only really care if the field is disabled or not based on a given context.

framework/widget/src/main/java/org/apache/ofbiz/widget/model/ModelFormField.java
```diff
@@ -2738,25 +2745,30 @@ public ModelForm getModelGrid(Map<String, Object> context) {
      */
     public static class HiddenField extends FieldInfo {
         private final FlexibleStringExpander value;
+        private final FlexibleStringExpander disabled;
```
danwatford, line 2748: Is the disabled property relevant for hidden fields? If not then this is a chance to remove a few lines of code.

danwatford (changes requested):

framework/widget/dtd/widget-form.xsd
```diff
@@ -857,7 +857,13 @@ under the License.
                     </xs:documentation>
                 </xs:annotation>
             </xs:attribute>
-            <xs:attribute name="disabled" type="xs:boolean" default="false"/>
+            <xs:attribute type="xs:string" name="disabled">
+                <xs:annotation>
+                    <xs:documentation>
+                        Set a field as disabled : not mutable, not focusable, not submitted if in a form.
+                    </xs:documentation>
```
danwatford, line 866: Could be worth including something about using expressions, similar to that the for `id-name` attribute:
```
Accepts ${} notation to allow use of expressions. Expressions must evaluate to the strings 'true', 'false' (without quotes) or empty (equivalent to false)
```

framework/widget/src/main/java/org/apache/ofbiz/widget/model/ModelFormField.java
florianMo: It's not really common but you can imagine to have a `<input type="hidden" disabled />`, which is valid I think, according to the spec (this is why I added this field on `HiddenField`, to be complete). It renders as an invisible and not submitted field. Not very useful (though I can imagine use cases where this could be handy), but since you can do it in HTML I thought we should support it.

framework/widget/src/main/java/org/apache/ofbiz/widget/model/ModelFormField.java
florianMo: done @danwatford

framework/widget/src/main/java/org/apache/ofbiz/widget/model/ModelFormField.java
florianMo: done

framework/widget/dtd/widget-form.xsd
florianMo: done

framework/widget/src/main/java/org/apache/ofbiz/widget/model/ModelFormField.java
```diff
@@ -2799,6 +2811,26 @@ public String getValue(Map<String, Object> context) {
             return getModelFormField().getEntry(context);
         }
 
+        /**
+         * Gets disabled.
+         * @return the disabled
+         */
+        public FlexibleStringExpander getDisabled() {
```
danwatford: Can this be renamed `getDisabledSpec` to differentiate it from calling `getDisabled` with context arguments. (Similar to the change made to the `ModelFormField` class above.)

Does anything use this method? If not, we should remove it or make it private.

framework/widget/src/main/java/org/apache/ofbiz/widget/model/ModelFormField.java
```diff
@@ -2799,6 +2811,26 @@ public String getValue(Map<String, Object> context) {
...
+            return disabled;
+        }
+
+        /**
+         *
+         * @param context the context
+         * @return evaluated value
+         */
+        public String getDisabled(Map<String, Object> context) {
```
danwatford: Should return a boolean based on whether the string evaluates to 'true', 'false', or empty (which would also be false).

framework/widget/src/main/java/org/apache/ofbiz/widget/model/ModelFormFieldBuilder.java
```diff
@@ -729,7 +729,7 @@ public String getConditionGroup() {
      * Gets disabled.*
      * @return the disabled
      */
-    public boolean getDisabled() {
+    public FlexibleStringExpander getDisabled() {
```
danwatford: Rename to `getDisabledSpec` to indicate it is no longer returning the disabled status of a field.

framework/widget/src/main/java/org/apache/ofbiz/widget/renderer/macro/RenderableFtlFormElementsBuilder.java
```diff
@@ -287,7 +287,7 @@ public RenderableFtl textField(final Map<String, Object> context, final ModelFor
                 .stringParameter("id", id)
                 .stringParameter("event", event != null ? event : "")
                 .stringParameter("action", action != null ? action : "")
-                .booleanParameter("disabled", disabled)
+                .stringParameter("disabled", String.valueOf(disabled))
```
danwatford: I don't think this change is needed. FTL is able to accept boolean parameters. 
One of the ideas behind RenderableFtl's parameters was to avoid encoding meaning in strings when other more focussed types, such as boolean an int, would do a better job.

framework/widget/src/main/java/org/apache/ofbiz/widget/model/ModelFormField.java
danwatford: ok

danwatford (changes requested):
A few changes requested around exposing disabled boolean state vs the specification of the disabled field attribute.  These changes might have knock-on effects to callers which would also need clean up.

framework/widget/src/main/java/org/apache/ofbiz/widget/model/ModelFormField.java
florianMo: right sorry I should have done the same here !

framework/widget/src/main/java/org/apache/ofbiz/widget/renderer/macro/RenderableFtlFormElementsBuilder.java
florianMo: I'll switch to a `boolean` in `HtmlFormMacroLibrary.ftl`, it's a little bit cleaner indeed, so this change is no longer needed (vs comparing a `string` to `true` in the macros).

framework/widget/src/main/java/org/apache/ofbiz/widget/model/ModelFormFieldBuilder.java
florianMo: > A few changes requested around exposing disabled boolean state vs the specification of the disabled field attribute. These changes might have knock-on effects to callers which would also need clean up.

Changes made @danwatford, so now we're passing a boolean to HTML macros (most notable adjustment).

danwatford (approved):

danwatford:
Tested as PASSED.

Tested by changing form `GISetupForms.xml#PartyAcctgPreference` and set the disabled attribute on the `errorGlJournalId` field defined on line 376. Changes were observed in the browser at page https://localhost:8443/accounting/control/PartyAcctgPreference

Added attribute to field:
```
disabled="${partyAcctgPreference.refundPaymentMethodId == 'ABN_CHECKING'}"
```

Then cycled through the values of Refund Payment Method ID on the form, saving after each change and observing whether the Error GL Journal ID drop down was disabled.

PR #563, opened 2022-11-17
Fixed: Synchronise FTL macro parameters for the various output types (OFBIZ-12126)

danwatford:
Following changes to how FTL macro parameter lists for HTML screens, corresponding changes needed to be made for text, xml, pdf (fo), csv and xls screens.

Tested using example and Layout demo screens, exercising exporting of screens to other formats
https://localhost:8443/example/control/main
https://localhost:8443/webtools/control/WebtoolsLayoutDemo


danwatford:
These changes are similar to those carried out under https://issues.apache.org/jira/browse/OFBIZ-6461

PR #574, opened 2022-12-27
Fixed: Add missing CSS for tree select in Helveticus

florianMo:
Fixed: Add missing CSS, so tree select can be properly used. Make expand/collapse links visible

(OFBIZ-12725)

https://issues.apache.org/jira/browse/OFBIZ-12725


priyasharma1 (approved):

nmalin:
Thanks Florian, I loaded your commit on trunk and branche 22.01, thanks for your PR

PR #577, opened 2023-01-09
Fixed: Permission check when creating a Product Facility (OFBIZ-12739)

danwatford:
Typo in the action being checked when creating a product facility prevented users with the FACILITY_CREATE permission, but not the CATALOG_CREATE permission, from creating Product Facilities.


danwatford:
Tested by creating a new user in an environment with demo data loaded. User has following permissions via a Security Group:

- OFBTOOLS_VIEW
- FACILITY_VIEW
- FACILITY_UPDATE
- FACILITY_CREATE
- CATALOG_VIEW

Successfully created a Product Facility for product WF-9943, meaning the FACILITY_CREATE permission was successfully used as a fallback for the absence of the CATALOG_CREATE permission.

PR #592, opened 2023-02-20
Improved: Improvement to createPartyRelationship service (OFBIZ-12765)

cshan-ecomify:
Improved: Improvement to createPartyRelationship service (OFBIZ-12765)

The createPartyRelationship service now allows new relationships of the same parties if there are no conflicts in their period of validity.

### DIFF
--- a/applications/party/groovyScripts/party/PartyServices.groovy
+++ b/applications/party/groovyScripts/party/PartyServices.groovy
@@ -427,13 +427,49 @@ def createPartyRelationship() {
                partyIdTo: parameters.partyIdTo,
                roleTypeIdFrom: parameters.roleTypeIdFrom,
                roleTypeIdTo: parameters.roleTypeIdTo)
-        .filterByDate()
         .queryList()
-    if (!partyRels) {
-        GenericValue partyRelationship = makeValue('PartyRelationship', parameters)
-        partyRelationship.create()
+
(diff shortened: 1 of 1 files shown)


applications/party/groovyScripts/party/PartyServices.groovy
```diff
@@ -427,13 +427,49 @@ def createPartyRelationship() {
...
+
+        // checks if there are any conflicts with the time periods of the entries
+        if (!parameters.thruDate) {
+            for(GenericValue partyRel : partyRels) {
+                if ((parameters.fromDate.before(partyRel.fromDate)) || (parameters.fromDate.before(partyRel.thruDate))) {
+                    return success()
+                }
+            }
+        } else {
+            EntityCondition partyCond =
```
dixitdeepak, line 449: Instead of having logic in createPartyRelationship service, we can make it simple create service, service will simply create party relationship record, and at the business side we can implement checks

mbrohl: Hi Deepak, this change is explixitely provided to enhance the business logic on this central point so that code using it does not have to deal with it.

mbrohl: @dixitdeepak are you ok with the change after my explanation?

PR #597, opened 2023-03-01
Improved: Ported createInventoryItem service to groovy (OFBIZ-12174)

danwatford:
Changes made as part of porting InventoryServices implementation from minilang to groovy.

Thanks: Sebastian Berg for implementation


danwatford:
Tested with demo data by receiving inventory into the WebStoreWarehouse facility.
Received inventory via Purchase Order and via product selection.
Manually altered the product's settings around entering Lots to test error cases in groovy implementation.

PR #621, opened 2023-04-13
Implemented: Use node/npm during gradle build from multiple sub-projects (OFBIZ-12789)

danwatford:
Added gradle plugin, ofbiz-node-conventions, to set conventional use of the gradle node plugin across OFBiz sub-projects (i.e. components).

Modified common-theme to install and cleanup NPM modules at appropriate points in the gradle task lifecycle.

This change will allow multiple components, including plugsin, to leverage the gradle node plugin in their build process without needing updates to the parent build.gradle script.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -27,4 +27,5 @@ Dockerfile
 /plugins/.svn
 /plugins/.github
 /plugins/.git
-
+/buildSrc/.gradle
+/buildSrc/build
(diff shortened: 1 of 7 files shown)


danwatford:
New node/npm build functionality for components can be observed by building the example plugin in PR https://github.com/apache/ofbiz-plugins/pull/77. 

A new sub-menu has been added to the Example plugin which incorporates a react web app, build using npm.

JacquesLeRoux:
Hi Daniel,

It's not OFBIZ-12787 but OFBIZ-12789 ;)

danwatford:
> Hi Daniel,
> 
> It's not OFBIZ-12787 but OFBIZ-12789 ;)

Good catch - I'll recreate the commit and force push to get this PR updated

PR #633, opened 2023-05-05
Fixed: Updates installation instructions to use JDK 17 instead of 11

mbrohl:


blling:
It is still use `JDK 11` in `README`, it also needs to change to `JDK 17`

JacquesLeRoux:
Thanks @blling , I'll handle it :)

JacquesLeRoux:
Done

PR #634, opened 2023-05-05
Improved: getConfigParam can't be configured at runtime (OFBIZ-12815)

thahn27:
In for example InterfaceConfig there are several methods which lead to UtilProperties instead of EntityUtilProperties and therefore cannot be configured during runtime.

Improved: In for example InterfaceConfig there are several methods which lead to UtilProperties instead of EntityUtilProperties and therefore cannot be configured during runtime. All of these methods don't access the database at all and just reference the UtilProperties methods. I rewrote the methods in EntityUtilProperties so first the database gets checked and if no entry in the database is found the fallback to UtilProperties happens. I left in all the current used methods and just placed my rewritten methods above the current used ones.

### DIFF
--- a/framework/entity/src/main/java/org/apache/ofbiz/entity/util/EntityUtilProperties.java
+++ b/framework/entity/src/main/java/org/apache/ofbiz/entity/util/EntityUtilProperties.java
@@ -123,6 +123,117 @@ public final class EntityUtilProperties implements Serializable {
         }
     }
 
+    public static <T> Object getPropertyValue(String resource, String name, Object defaultValue, Delegator delegator, Class<T> clazz) {
+        Map<String, String> propMap = getSystemPropertyValue(resource, name, delegator);
+        if ("Y".equals(propMap.get("isExistInDb"))) {
+            String s = propMap.get("value");
+            if (UtilValidate.isEmpty(s)) {
+                return defaultValue;
+            }
+            try {
+                if (clazz.componentType() == Integer.class) {
+                    return Integer.valueOf(s);
+                }
+                if (clazz.componentType() == Double.class) {
+                    return Double.valueOf(s);
+                }
+                if (clazz.componentType() == Long.class) {
+                    return Long.valueOf(s);
+                }
+                if (clazz.componentType() == Float.class) {
+                    return Float.valueOf(s);
+                }
+                if (clazz.componentType() == BigInteger.class) {
+                    return Float.valueOf(s);
+                }
+                if (clazz.componentType() == BigDecimal.class) {
+                    return Float.valueOf(s);
+                }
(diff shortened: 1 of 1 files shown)


framework/entity/src/main/java/org/apache/ofbiz/entity/util/EntityUtilProperties.java
```diff
@@ -123,6 +123,117 @@ public static String getPropertyValueFromDelegatorName(String resource, String n
...
 
+    public static <T> Object getPropertyValue(String resource, String name, Object defaultValue, Delegator delegator, Class<T> clazz) {
+        Map<String, String> propMap = getSystemPropertyValue(resource, name, delegator);
+        if ("Y".equals(propMap.get("isExistInDb"))) {
+            String s = propMap.get("value");
+            if (UtilValidate.isEmpty(s)) {
+                return defaultValue;
+            }
+            try {
+                if (clazz.componentType() == Integer.class) {
```
gilPts, line 134: Hello, since it is for trunk and java 17 is available, what do you think about using switch pattern matching (and Guarded Pattern) 
ref : https://openjdk.org/jeps/406
Thanks and regards, 
Gil

gilPts: Mmm, i got carried away discovering those new switch syntax :), simple switch are sufficient here, sorry.

thahn27: Hey Gil, 

i closed this PR and opened another one because the commit description was wrong. The new PR: https://github.com/apache/ofbiz-framework/pull/635
Also i will adjust the existing code with a switch in an upcoming commit in the new PR.

Thanks and regards,
Tobias

gilPts: Thanks to you Tobias :)

thahn27: Hey @gilPts ,

i tried to implement the switch statement instead of the if statements and i can't really make it work. The pattern matching documentation in the ref you provided works with objects which have been instantiated before. Maybe stick to the if statements? 
Maybe you could provide an example or explain how you would implement such switch statement.

Thanks and regards,
Tobias

gilPts: Yes pattern matching is not applicable here, hence my second comment, sorry for the misunderstanding.  If in those cases are fine, since we are operating on `Class` object test...
Thanks

PR #635, opened 2023-05-08
Improved: EntityUtil getProperty Methods dont use entity (OFBIZ-12815)

thahn27:
The getProperty methods in EntityUtilProperties don't use entity at all. All of the getProperty methods simply lead to UtilProperties and therefore no configure during runtime is possible.

Improved: New methods have been written so the entity usage is now functional.

### DIFF
--- a/framework/entity/src/main/java/org/apache/ofbiz/entity/util/EntityUtilProperties.java
+++ b/framework/entity/src/main/java/org/apache/ofbiz/entity/util/EntityUtilProperties.java
@@ -123,6 +123,117 @@ public final class EntityUtilProperties implements Serializable {
         }
     }
 
+    public static <T> Object getPropertyValue(String resource, String name, Object defaultValue, Delegator delegator, Class<T> clazz) {
+        Map<String, String> propMap = getSystemPropertyValue(resource, name, delegator);
+        if ("Y".equals(propMap.get("isExistInDb"))) {
+            String s = propMap.get("value");
+            if (UtilValidate.isEmpty(s)) {
+                return defaultValue;
+            }
+            try {
+                if (clazz.componentType() == Integer.class) {
+                    return Integer.valueOf(s);
+                }
+                if (clazz.componentType() == Double.class) {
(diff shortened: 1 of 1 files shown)


JacquesLeRoux:
This sounds good to me. 

Should we not deprecate the previous `getPropertyNumber(String resource, String name, double defaultValue)` until `getPropertyAsBigDecimal(String resource, String name, BigDecimal defaultNumber)` ?

thahn27:
Hey Jacques,

I will provide another commit in which i adjust the switch statement mentioned by Gil and also deprecate the previous ones.

JacquesLeRoux:
Hi @thahn27,

I checked the usage in Java and Groovy code of the previous methods I talked about above. Most of them are not used at all OOTB.
getPropertyAsBoolean and getPropertyAsBigDecimal are both used twice in Java, getPropertyAsInteger is used 21 times in Java and 6 in Groovy.

It could be that OFBiz users are using them more. But it's really easy to replace them by UtilProperties ones or the new EntityUtilProperties ones. So maybe we could remove them all together OOTB. And replace those used by calls to the new ones you introduce. I'll start a discussion on dev ML about that.

JacquesLeRoux:
I send an email to dev ML 1 hour ago but it does not appear yet. I wonder why, but let's wait

framework/entity/src/main/java/org/apache/ofbiz/entity/util/EntityUtilProperties.java
```diff
@@ -123,6 +123,117 @@ public static String getPropertyValueFromDelegatorName(String resource, String n
         }
     }
 
+    public static <T> Object getPropertyValue(String resource, String name, Object defaultValue, Delegator delegator, Class<T> clazz) {
+        Map<String, String> propMap = getSystemPropertyValue(resource, name, delegator);
```
dixitdeepak, line 127: Can we use ObjectType.simpleTypeConvert method ?

PR #639, opened 2023-05-15
Improvement of the Check xml tag (OFBIZ-12821)

MaximeFlame:
Added features :
- Pre-select / Pre-check
- -> Chose which should be pre-check
- Pre-unSelect / Pre-unCheck
- -> All checked and select whose should not be checked

EDIT : 

[OFBIZ-12821](https://issues.apache.org/jira/browse/OFBIZ-12821)

Thanks:
Néréide

### DIFF
--- a/applications/order/widget/ordermgr/QuoteForms.xml
+++ b/applications/order/widget/ordermgr/QuoteForms.xml
@@ -29,7 +29,7 @@ under the License.
         <auto-fields-entity entity-name="Quote" default-field-type="find"/>
         <field name="quoteId" title="${uiLabelMap.OrderOrderQuoteId}"></field>
         <field name="quoteTypeId" title="${uiLabelMap.OrderOrderQuoteTypeId}">
-            <check all-checked="false" unchecked-by-default="quoteTypeId">
+            <check all-checked="false">
                 <entity-options entity-name="QuoteType" key-field-name="quoteTypeId"/>
             </check>
         </field>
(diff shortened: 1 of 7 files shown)


nmalin:
Hi @MaximeFlame,

Xml attributes checked-by-default and unchecked-by-default seems to add only code complexity without most value.

With the possibility to check form a pk list, you have natively this functionality through the prepare statement.

MaximeFlame:
Code updated @nmalin

JacquesLeRoux:
Hi @MaximeFlame,

Do your code changes answer to @nmalin's suggestion? BTW I don't clearly understand your suggestion Nicolas :)
Also there is a test issue...

nmalin:
Hi @JacquesLeRoux, I shared with him on IRL,

I said we can have the same functionnality to precheck or uncheck easly with set the good id list on the context. So  adding two attributes to do that is redundant and finally add unnecessary code.

JacquesLeRoux:
Nicolas, do you mean in the screen or java methods? Maybe we could document that. It's still not crystal clear to me, sorry

MaximeFlame:
> Hi @MaximeFlame,
> 
> Do your code changes answer to @nmalin's suggestion? BTW I don't clearly understand your suggestion Nicolas :) Also there is a test issue...

Yes

JacquesLeRoux:
Hi @nmalin, 

Thanks for the explanation, I got it looking at QuoteForms.xml where OTHER_QUOTE and PROPOSAL are preset.
I simply amended the doc in widget-form.xsd. I'll commit from my machine

PR #644, opened 2023-06-15
Improved: DataResourceWorker: New method getMimeType(...) (OFBIZ-9403)

cshan-ecomify:
Improved: DataResourceWorker: New method getMimeTypeFromFileName(...) with extracted (refactored) code from getMimeType(...) method (OFBIZ-9403)

The method DataResourceWorker.getMimeType has been refactored into two separate methods.


mbrohl:
@cshan-ecomify I think we should take the defaultMimeType from applications/content/config/content.properties instead of a hard coded one, can you please change it accordingly?

PR #648, opened 2023-06-30
Improved: Use a version catalog

dtrunk90:
Using a version catalog makes things easier:
- You can update a single version for multiple artifacts (e.g. Tomcat)
- You have all the artifacts and their versions at a single place separated from other stuff

See also https://docs.gradle.org/current/userguide/platforms.html

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -29,14 +29,15 @@ plugins {
     id 'checkstyle'
     id 'codenarc'
     id 'maven-publish'
-    id 'org.asciidoctor.jvm.convert' version '3.3.2' // 4.0.2 does not compile
-    id 'org.asciidoctor.jvm.pdf' version '3.3.2'     // 4.0.2 does not compile
-    id 'org.owasp.dependencycheck' version '9.0.9' apply false //Not tested after 7.4.4
-    id 'se.patrikerdes.use-latest-versions' version '0.2.18' apply false
-    id 'com.github.ben-manes.versions' version '0.51.0' apply false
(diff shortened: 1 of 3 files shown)


JacquesLeRoux:
Hi @dtrunk90, 

I remember having seen that before, but completly forgot about it. 
Could you please update the PR, or create anew. At 1st glance, the conflicts seem a bit hard to resolve

TIA

dtrunk90:
Hi @JacquesLeRoux,

it took some hours but it's finally done.

mbrohl:
@dtrunk90 Can you elaborate a bit more what advantages this approach brings?
At first glance it seems to make things more complicate?

JacquesLeRoux:
Hi @dtrunk90, @mbrohl ,

I must say last time I just had a glance, notably on the Gradle Page. After reading a bit more there, with a glance as the resulting code here, I indeed wonder if we need that OOTB. 

@dtrunk90, I suggest to start a thread in the dev ML (dev@ofbiz.apache.org). That's what we do when we hesitate and want to make a community decision. Just few explaining words should be enough.

JacquesLeRoux:
As a note, here is the Gradle page summary for the catalog feature:

> In summary, using a catalog is always a good engineering practice as it centralizes common definitions, allows sharing of dependency versions or plugin versions, but it is an "implementation detail" of the build: it will not be visible to consumers and unused elements of a catalog are just ignored.

Again, not sure we need that, we need to deal with it in depth.

PR #652, opened 2023-07-24
Improved: Refactoring groovy folder structure and package declaration (OFBIZ-12813)

wpaetzold:


applications/accounting/src/main/groovy/org/apache/ofbiz/admin/AcctgAdminServices.groovy
```diff
@@ -15,7 +15,10 @@
  * KIND, either express or implied.  See the License for the
  * specific language governing permissions and limitations
  * under the License.
- */
+*/
+
+package org.apache.ofbiz.admin
```
dixitdeepak, line 20: Package name should be org.apache.ofbiz.accounting.admin, Like we have package hierarchy for java.

mbrohl:
Closed in favor of https://github.com/apache/ofbiz-framework/pull/653

PR #654, opened 2023-07-26
Improved: ContentWrapper empty string result breaks simple FTL null check and default syntax  (OFBIZ-10194)

Adrian-Wolf:
Improved: ContentWrapper empty string result breaks simple FTL null check and default syntax (OFBIZ-10194)


mbrohl:
This looks good to me an is field tested in custom projects for some time now.
Any objections to merge?

PR #655, opened 2023-07-26
Implemented: Extending calculateProductPrice (custom) Service and ShoppingCart / OrderItem to deliver/hold an individual discountRate value (OFBIZ-12802)

Adrian-Wolf:
Implemented: Extending calculateProductPrice (custom) Service and ShoppingCart / OrderItem to deliver/hold an individual discountRate value (OFBIZ-12802)

+ Addition of additional context information "productStoreGroupId" and "partyId" to the call parameters of a "customMethodName" price-calc-service
+ Obtaining the fields discountRate and listPrice from calling a "customMethodName" price-calc-service

+ discountRate as Result Field
+ Transport original discount rate from custom price calculation over cart to order to have the correct rate instead of back-calculating it from unitPrice and unitListPrice


mbrohl:
Looks good to me. Any objections to merge?

PR #673, opened 2023-10-13
Implemented: Add docker build functionality to release18.12 branch (OFBIZ-12856)

danwatford:
Added docker build sources, documentation and example usage to the relese18.12 branch. Also added GitHub Actions workflow to build and push container images to the GitHub Container Registry (ghcr.io).


JacquesLeRoux:
Thanks Daniel,

I did not get into all details, but as it works for trunk and next demos I see no reason why it would not work for stable
+1 for pushing

PR #674, opened 2023-10-30
Fixed Issue with OFBiz Job Scheduler and Daylight Saving Time  (OFBIZ-12864)

dixitdeepak:
Fixed: Issue with OFBiz Job Scheduler and Daylight Saving Time
OFBIZ-12864

Explanation

The issue occurs when DST changes, and OFBiz fails to schedule recurring jobs properly. This is due to a condition in the PersistedServiceJob.createRecurrence method where it compares the next scheduled time (next) with the start time (startTime) for the job.

To address the issue, I've proposed adding a new field named JobSandbox.runTimeEpoch. This field would store the UTC format epoch milliseconds of the runtime date. When scheduling or rescheduling recurring jobs, the system would use the UTC epoch stored in JobSandbox.runTimeEpoch for comparison.

This solution ensures that the system uses a consistent, UTC-based time for scheduling and rescheduling recurring jobs, even when DST changes affect the local time.

To implement this solution, 
- Modify the PersistedServiceJob.createRecurrence method to calculate and store the UTC epoch milliseconds in the JobSandbox.runTimeEpoch field.
- Update the code responsible for polling and rescheduling jobs to use the JobSandbox.runTimeEpoch field when it is set. If the field is not set, system would fall back to getting the runtime date to filter the jobs.

By using this approach, the system should be able to handle recurring job scheduling more reliably, especially when DST changes are involved, as it ensures that all time comparisons are made in a consistent UTC format.

Thanks @jacopoc  for the valuable discussion and suggestion!!

### DIFF
--- a/framework/service/entitydef/entitymodel.xml
+++ b/framework/service/entitydef/entitymodel.xml
@@ -44,6 +44,7 @@ under the License.
         <field name="jobId" type="id"></field>
         <field name="jobName" type="name"></field>
         <field name="runTime" type="date-time"></field>
+        <field name="runTimeEpoch" type="numeric"/><!-- Store the Epoch millisecond to avoid DST change issue (when clock set 1 hr back) -->
         <field name="priority" type="numeric"></field>
         <field name="poolId" type="name"></field>
         <field name="statusId" type="id"></field>
--- a/framework/service/src/main/java/org/apache/ofbiz/service/job/JobManager.java
+++ b/framework/service/src/main/java/org/apache/ofbiz/service/job/JobManager.java
@@ -185,20 +185,23 @@ public final class JobManager {
             By adding the runTimeEpoch field to handle DST changes in recurring job scheduling is a practical solution.
             By storing the runtime epoch in UTC format, to make the system DST-aware, which helps prevent issues related to time changes,
                 especially when the clock is set back by 1 hour during the transition.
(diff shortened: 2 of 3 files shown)


JacquesLeRoux:
Hi Deepak,

It's something "special" to test, but I can't see any side affects by adding the lines as your did => +1

mbrohl:
+1 would be good if this could be squashed into 1 commit.
Thanks Deepak!

JacquesLeRoux:
We have by default the Squash and merge button ;)

dixitdeepak:
> Hi Deepak,
> 
> It's something "special" to test, but I can't see any side affects by adding the lines as your did => +1

Hi @JacquesLeRoux 

## Steps to Reproduce the Issue

1. Set the `ofbiz.timeZone.default` to `America/New_York` (You can use any time zone that follows Daylight Saving Time).
2. Set the system time to November 5, 2023, at 01:00 AM. (In 2023, DST ends on November 5th at 2:00 AM, at this moment clock is set back by 1 hour.)
3. Schedule a job (e.g., the `ping` service) at various frequencies, such as 15 minutes, 30 minutes, and 1 hour.
4. If the system time is 1:45 AM (EDT) and the job has a recurring frequency of 30 minutes, the next job time will be scheduled at 1:15 AM (EST) as DST ends at 2:00, causing the system clock to reset back by 1 hour.
5. The Job Poller will pick up this job at the next polling, but it fails to reschedule because the next start time is not greater than the current job start time.

PR #682, opened 2024-01-04
Removed unnecessary steps when building docker image with plugins (OFBIZ-12876)

danwatford:


JacquesLeRoux:
Thanks Daniel, sounds good to me +1

PR #684, opened 2024-01-23
Fixed: Keep needed classpath entries in the Eclipse .classpath file (OFBIZ-12880)

mbrohl:
Keeps classpath entries for config folders and dtds which are removed from the .classpath file after it is generated by the Eclipse Gradle task.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -411,14 +411,12 @@ eclipse.classpath.file.whenMerged { classpath ->
 
         classpath.entries.removeAll { entry ->
             // remove any "src" entries in .classpath of the form /componentName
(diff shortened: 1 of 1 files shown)


JacquesLeRoux:
Hi Michael, this reverts https://issues.apache.org/jira/browse/OFBIZ-11071 is that normal? If yes could you please give us more information on the why?

TIA

mbrohl:
Hi @JacquesLeRoux ,

the exclusion to not remove "framework/base/config" and "framework/base/dtd" is not necessary anymore when we do not remove /config and /dtd folders at all.

JacquesLeRoux:
OK, I see thanks

PR #687, opened 2024-02-04
Fixed: Gradle eclipse target produces duplicate build path entries in .classpath (OFBIZ-12888)

mbrohl:
Removes the groovyScripts source set, the compileGroovyScriptsGroovy task configuration and changes the pre-push hook to "checkstyleMain codenarcMain codenarcTest" instead of "checkstyleMain codenarcGroovyScripts".

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -352,20 +352,6 @@ sourceSets {
             srcDirs = getDirectoryInActiveComponentsIfExists('src/test/resources')
         }
     }
-
-    groovyScripts {
-        groovy {
-            srcDirs += getDirectoryInActiveComponentsIfExists('src/main/groovy')
-            compileClasspath += sourceSets.main.compileClasspath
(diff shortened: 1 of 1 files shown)


JacquesLeRoux:
It's only a matter of performance for the hook. I'd like to have @danwatford's opinon about removing the compileGroovyScriptsGroovy Gradle task...

danwatford:
> It's only a matter of performance for the hook. I'd like to have @danwatford's opinon about removing the compileGroovyScriptsGroovy Gradle task...

Yes - the groovyScripts source set should be removed along with the changes to the compileGroovyScriptsGroovy task. 

This PR should be merged, with corresponding changes to the OFBiz buildbot definition to prevent build failure due to the absence of the build/reports/codenarc/groovyScripts.html file. Explanation given in comments here - https://issues.apache.org/jira/browse/OFBIZ-12888.

PR #693, opened 2024-02-15
OFBIZ-12900 Have screen FindPartyClassificationGroups show Party MainActionMenu

PierreSmits:
Currently the FindPartyClassificationGroups screen in PartyClassificationScreens.xml does not show the MainActionMenu of the party component. For a consistent user experience this should be.

modified: PartyClassificationsScreens.xml
- added decorator-section 'pre-body', having MainActionMenu


JacquesLeRoux:
Hi Pierre,

It' s not the real commit title we expect (type missing), but it's OK with me (because it's really trivial)

PierreSmits:
Hi Jacques,

Thanks for the feedback, but lets not dwell on trivialities.

PR #698, opened 2024-02-17
Fixed: PaymentGroup Cancell button is shown to viewer (OFBIZ-12820)

PierreSmits:
Currently when a user with only view permissions accesses the PaymentGroup overview, as demonstrated in demo-trunk with userid=auditor, the action trigger to cancel a PaymentGroup is shown. See attached image.

modified: PaymentGroupForms.xml
- removed action trigger field regarding cancelCheckRunPayments

### DIFF
--- a/applications/accounting/widget/PaymentGroupForms.xml
+++ b/applications/accounting/widget/PaymentGroupForms.xml
@@ -61,11 +61,6 @@ under the License.
                 <parameter param-name="paymentGroupId"/>
             </hyperlink>
         </field>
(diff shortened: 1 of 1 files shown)


mbrohl:
It looks like the action is simple removed for everyone instead of checking the permission?

PierreSmits:
The trigger to cancel a PaymentGroup is still available on the page of the PaymentGroup. See https://demo-trunk.ofbiz.apache.org/accounting/control/PaymentGroupOverview?paymentGroupId=9000.

IMO, cancelling a PaymentGroup in an overview with less detail than available in the page of the PaymentGroup should be discouraged as accidental cancellations do happen.

PR #711, opened 2024-02-27
Improved: Move from hard-code menu location to parameterized (OFBIZ-12918)

PierreSmits:
Customization is a key aspect of the OFBiz (Open For Business) platform, providing businesses with a robust and adaptable base to customize their enterprise software solutions according to their unique operational requirements. The *Menus.xml file of a component, essential for designing user interfaces that are both intuitive and tailored to the specific roles and tasks of its users, plays a significant role in this context, as referenced within screen widgets and Freemarker templates. However, as customizations and the project's source code evolve over time, the likelihood of encountering merge conflicts increases significantly, especially when attempting to incorporate customizations alongside ongoing bug fixes and enhancements from the project. To improve the appeal of OFbiz and the developer experience, it is recommended to shift from fixed menu locations to a parameterized approach in our components and plugins. This change would enable developers to more seamlessly incorporate a 'custom' menu, facilitating the integration of project updates.

modified:
- accounting/WEB-INF/web.xml: added parameter for mainMenuLocation 
- various *Screens.xml files: changing location of 'include-menu' elements from location="component://accounting/widget/AccountingMenus.xml to location="${parameters.mainMenuLocation}"

### DIFF
--- a/applications/accounting/webapp/accounting/WEB-INF/web.xml
+++ b/applications/accounting/webapp/accounting/WEB-INF/web.xml
@@ -43,6 +43,11 @@ under the License.
         <param-name>partyDecoratorLocation</param-name>
         <param-value>component://accounting/widget/CommonScreens.xml</param-value>
     </context-param>
+    <context-param>
+        <description>The location of the main *Menus XML file to be used in this webapp; referred to as a context variable in screen def XML files.</description>
+        <param-name>mainMenuLocation</param-name>
+        <param-value>component://accounting/widget/AccountingMenus.xml</param-value>
+    </context-param>
 
     <filter>
         <display-name>ControlFilter</display-name>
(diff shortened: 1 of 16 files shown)


applications/accounting/webapp/accounting/WEB-INF/web.xml
```diff
@@ -43,6 +43,11 @@ under the License.
         <param-name>partyDecoratorLocation</param-name>
         <param-value>component://accounting/widget/CommonScreens.xml</param-value>
     </context-param>
+    <context-param>
```
MkLeila, line 46: Hello Pierre,
I think the same thing should be done also in ar and ap web.xml (they use screens such as invoiceOverview from accounting and will be impacted)

PierreSmits: HI @MkLeila,

See my reply in the ticket.

PR #713, opened 2024-02-28
Fixed: Remove rolled-over log files (OFBIZ-12897)

danwatford:
Changed file naming pattern used for rolled over log files in the OFBiz default log4j2.xml configuration file such that the date is not added to rolled log file names.

Removing the date component from log file names means the RollingFile log appender will correctly remove older log files when the maximum number of rolled log files is reached.


JacquesLeRoux:
Thanks Daniel,

That sounds good to me as we have the dates in log files anyway

PR #714, opened 2024-02-28
Improved: Have library dependencies moved to a dependencies.gradle file (OFBIZ-10924)

PierreSmits:
Currently the libraries needed by ofbiz are defined in the build.gradle file. These should reside in a separate dependencies.gradle file that is referenced in the build.gradle file, like the common.gradle. As is common practice in other projects/solutions that work with dependencies on external libraries.

modified:
- build.gradle: removed implementation, testImplementation and runtimeOnly library dependencies added:
dependencies.gradle, having the implementation, testImplementation and runtimeOnly library dependencies


PierreSmits:
Closing due to library upgrade issue and merge-conflict.

PR #721, opened 2024-03-04
Improved: MFG-Move from hard-code menu location to parameterized (OFBIZ-12928)

PierreSmits:
Customization is a key aspect of the OFBiz (Open For Business) platform, providing businesses with a robust and adaptable base to customize their enterprise software solutions according to their unique operational requirements. The *Menus.xml file of a component, essential for designing user interfaces that are both intuitive and tailored to the specific roles and tasks of its users, plays a significant role in this context, as referenced within screen widgets and Freemarker templates. However, as customizations and the project's source code evolve over time, the likelihood of encountering merge conflicts increases significantly, especially when attempting to incorporate customizations alongside ongoing bug fixes and enhancements from the project. To improve the appeal of OFbiz and the developer experience, it is recommended to shift from fixed menu locations to a parameterized approach in our components and plugins. This change would enable developers to more seamlessly incorporate a 'custom' menu, facilitating the integration of project updates.

modified:
- manufacturing/WEB-INF/web.xml: added parameter for mainMenuLocation 
- various *Screens.xml files: changing location of 'include-menu' elements from location="component://manufacturing/widget/ManufacturingMenus.xml to location="${parameters.mainMenuLocation}"

### DIFF
--- a/applications/manufacturing/webapp/manufacturing/WEB-INF/web.xml
+++ b/applications/manufacturing/webapp/manufacturing/WEB-INF/web.xml
@@ -43,6 +43,11 @@ under the License.
         <param-name>commonManufacturingDecoratorLocation</param-name>
         <param-value>component://manufacturing/widget/manufacturing/CommonScreens.xml</param-value>
     </context-param>
+    <context-param>
+        <description>The location of the menus file to be used in this webapp; referred to as a context variable in screen def XML files.</description>
+        <param-name>mainMenuLocation</param-name>
+        <param-value>component://manufacturing/widget/manufacturing/ManufacturingMenus.xml</param-value>
+    </context-param>
 
     <filter>
         <display-name>ControlFilter</display-name>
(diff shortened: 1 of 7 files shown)


danwatford:
Hi @PierreSmits ,

The commit comments for some of your PRs are very similar and quite verbose, making it a bit difficult to tell the difference.

I think someone quickly scanning the commit log might not realise they are different sets of changes.

Since these PRs seem to be part of a wider programme of changes, would it be better to have a top-level jira ticket describing the nature of the change, and then sub-tasks which apply those changes to the various OFBiz applications?

This would allow you to have a line in a comment such as:
"Part of the menu parameterization error", followed by any application specific notes on the changes. This should also reduce the amount of comment needed for each of the commits which are part of the wider programme of work.

PierreSmits:
@danwatford 

Thank you for your feedback and suggestions. It is a pity that the code improvements won't go in.

JacquesLeRoux:
Indeed @danwatford , I agree and it's easy to change for this PR and followings

PierreSmits:
I will convert the ticket to a parent ticket, and submit PR's per sub-tasks there-under. However, given what is agreed upon some duplication of words in commit messages and Pull Request description cannot be avoided.

PR #722, opened 2024-03-04
Improved: Use RenderableFtlElements to render drop down fields (OFBIZ-12127)

danwatford:
Part of the OFBIZ-11456 refactoring effort.

Removed unused renderDropDownField FTL macro attributes from the various FormMacroLibrary.ftl files. Synchronised the attribute lists among the files and changed some attributes to have default values and/or use non-string types.

Moved logic used to construct the call to the renderDropDownField macro into RenderableFtlFormElementsBuilder where the logic now constructs a RenderableFtlMacroCall object rather than a long string.

Use of RenderableFtlMacroCall means objects, rather than string representations of objects, are passed to FTL. This has allowed removal of workarounds introduced in OFBIZ-6504 to escape characters passed to Freemarker in macro attributes.


danwatford:
A change to package-lock.json was accidently added to this PR. I'll remove it from the commit and push again.

PR #749, opened 2024-04-11
improved: have profileMenuLocation (OFBIZ-12981)

PierreSmits:
Profile screens are used in various components as portalpageportlets, and have references to various party menus. As menus in the various profile screens are parameterized to use ${parameters.mainMenuLocation} this generates errors in the components/applications where these screens are used as portal page elements.

Having a context-param profileMenuLocation resolves this issue.

modified:
partty - web.xml: have context-param profileMenuLocation
party - ProfileScreens.xml:  change location profile menus various screens

### DIFF
--- a/applications/order/webapp/ordermgr/WEB-INF/web.xml
+++ b/applications/order/webapp/ordermgr/WEB-INF/web.xml
@@ -43,6 +43,11 @@ under the License.
         <param-name>mainMenuLocation</param-name>
         <param-value>component://order/widget/ordermgr/OrderMenus.xml</param-value>
     </context-param>
+    <context-param>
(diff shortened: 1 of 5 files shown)


mbrohl:
We now have mixed changes for order and party and two Jiras together in one PR, this should be splitted.

I also do not understand those changes. The error seems to be introduced by the changes made by you to parametrize the menu location. Now it becomes necessary to make this even more complicated by introducing additional context parameters. The locations for mainMenuLocation and requestMenuLocation are the same, why are they necessary?

PierreSmits:
The requestMenuLocation  was necessary to be implemented due to the fact that request features are incorporated in the projectmgr plugin. Regarding the projectmgr menu parameterisations ticket OFBIZ-12975 has been created.

PR #779, opened 2024-04-21
Improved: HumanRes-Main page (OFBIZ-13050)

PierreSmits:
Current main page shows the OrgTree screen in a hard-coded way for about 50% of the available screen.

modified: CommonScreens.xml/screen Main
- added actions elements to work with PortalPage
- added include-portal-page element
- removed display conditio

### DIFF
--- a/applications/humanres/data/HumanResPortletData.xml
+++ b/applications/humanres/data/HumanResPortletData.xml
@@ -20,15 +20,27 @@
 <entity-engine-xml>
 
     <PortletCategory portletCategoryId="HUMANRES" description="Human Resource Portlets"/>
-    <PortalPortlet portalPortletId="myLeave"
-        portletName="My Leave"
-        screenName="MyLeaveList"
-        screenLocation="component://humanres/widget/EmployeeScreens.xml"
-        description="My Leaves form Human Resource" securityServiceName="portalPermissionIsEmployee" securityMainAction="VIEW"/>
+    <PortalPortlet portalPortletId="myLeave" portletName="My Leave"
+        description="My Leaves form Human Resource" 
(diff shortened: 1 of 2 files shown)


JacquesLeRoux:
Hi @PierreSmits ,

When applying your PR as patch, I get:

> java.lang.RuntimeException: Could not find PortalPage with portalPageId []

PierreSmits:
HI Jacques,

Did you clear your cache? Logging out and in again could also help.

JacquesLeRoux:
Hi Pierre,

Sorry, I simply forgot to import the data :/

JacquesLeRoux:
The screen separation is static. Did you think about the case where the tree would much grow? What will happen then? The need to see the leaves might be problematic. Even if a horizontal scroolbar appears. Did you test that? Because then you have to scrool. I think it's the reason of all this empty space.

PierreSmits:
HI Jacques,

Thank you for your review and feedback.
I can't tell why the screen separation is as it is. But if that is an issue: a new ticket is warranted. Testing the screen with leaves to the n-th power could be fun exercise (for someone...).
Let's try not to make a mountain out of an anthill now. If there ever has been (is) an adopter that had/has that issue (and he/she reported it to this project), I am confident that:

1.  that adopter can have an paid developer implementing a solution that fixes the issue.
2. if and when reported to the project, it can be addressed then via a new ticket.

But I doubt that the project will ever have to address such. I expect the given adopter to be a SMB enterprise with a hierarchy as flat as possible. Would love to see someone come forward via one of the mailings and say: I have (or work for) a company that uses the HR component of OFBiz for 10s of departments having 10s of sub-departments, with each having 10s of other organizational units thereunder.

PR #790, opened 2024-04-29
Improved: convert FinAccount getArithmeticSettingsInline to groovy (OFBIZ-13074)

PierreSmits:
modified:
- ofbiz-component.xml: added loader for services.xml
- FinAccountServices.xml removed simple-method getArithmeticSettingsInline, replaced ref in other simple-methods with new groovy service

added:
- services.xml: added groovy service getArithmeticSettingsFinAccount
- AccountingService.groovy: having getArithmeticSettingsFinAccount

### DIFF
--- a/applications/accounting/ofbiz-component.xml
+++ b/applications/accounting/ofbiz-component.xml
@@ -31,6 +31,8 @@ under the License.
     <entity-resource type="data" reader-name="seed" loader="main" location="data/AccountingPortletData.xml"/>
     <entity-resource type="data" reader-name="seed-initial" loader="main" location="data/AccountingScheduledServiceData.xml"/>
 
+    <service-resource type="model" loader="main" location="servicedef/services.xml"/>
+
     <!-- General Payment Service Definitions -->
     <service-resource type="model" loader="main" location="servicedef/services_agreement.xml"/>
     <service-resource type="model" loader="main" location="servicedef/services_admin.xml"/>
--- a/applications/accounting/src/main/groovy/org/apache/ofbiz/accounting/accounting/AccountingServices.groovy
+++ b/applications/accounting/src/main/groovy/org/apache/ofbiz/accounting/accounting/AccountingServices.groovy
@@ -21,7 +21,6 @@ package org.apache.ofbiz.accounting.accounting
 import org.apache.ofbiz.entity.util.EntityUtilProperties
 
 Map getArithmeticSettingsFinAccount() {
-    Map result = success()
     arithmeticSettings = [
         roundingDecimals: EntityUtilProperties.getPropertyValue('arithmetic', 'finaccount.decimals', '2', delegator),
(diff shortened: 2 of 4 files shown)


applications/accounting/src/main/groovy/org/apache/ofbiz/accounting/accounting/AccountingServices.groovy
```diff
@@ -0,0 +1,32 @@
...
+
+Map getArithmeticSettingsFinAccount() {
+    Map result = success()
+    arithmeticSettings = [:]
+    roundingDecimals = EntityUtilProperties.getPropertyValue('arithmetic', 'finaccount.decimals', '2', delegator)
+    arithmeticSettings.put('roundingDecimals', roundingDecimals)
+    roundingMode = EntityUtilProperties.getPropertyValue('arithmetic', 'finaccount.roundingSimpleMethod', 'roundingMode', 'HalfUp', delegator)
+    arithmeticSettings.put('roundingMode', roundingMode)
+    result.put('arithmeticSettings', arithmeticSettings)
+    return success (arithmeticSettings: arithmeticSettings)
```
nmalin: you can replace this by : 
```
Map arithmeticSettings = [roundingDecimals: EntityUtilProperties.getPropertyValue('arithmetic', 'finaccount.decimals', '2', delegator),
                                               roundingMode: EntityUtilProperties.getPropertyValue('arithmetic', 'finaccount.roundingSimpleMethod', 'roundingMode', 'HalfUp', delegator)]
return success([arithmeticSettings: arithmeticSettings])

File: applications/accounting/ofbiz-component.xml [hunk 2]
```diff
@@ -31,6 +31,8 @@ under the License.
     <entity-resource type="data" reader-name="seed" loader="main" location="data/AccountingPortletData.xml"/>
     <entity-resource type="data" reader-name="seed-initial" loader="main" location="data/AccountingScheduledServiceData.xml"/>
 
+    <service-resource type="model" loader="main" location="servicedef/services.xml"/>
```
nmalin, line 34: I suggest to use a current service file instead introduce a new one

applications/accounting/src/main/groovy/org/apache/ofbiz/accounting/accounting/AccountingServices.groovy
PierreSmits: Thanks Nicolas, for your time and effort.

I will improve this piece of code.

applications/accounting/ofbiz-component.xml
PierreSmits: Thanks Nicolas, for the suggestion.

However, there are more of those arithmetic functions to get such setting used in various xml, groovy and java files. I prefer to have those in a single services.xml file, instead having those spread across many files.

Plus, having a generic 'services.xml' in the accounting component brings the component more in line with others (and the templates for creating a new plugin). Such would enhance the developer (and thus the contributor) experience. 

I trust you find my solution acceptable.

PR #793, opened 2024-05-03
Fixed: PartyTypeAttr correction (OFBIZ-13076)

stschikin:
The entity PartyTypeAttr was missing an attrValue field and the description field needed to be renamed to align with other similar entities such as AgreementAttribute, AgreementItemAttribute, ContactMechAttribute, PartyAttribute

### DIFF
--- a/applications/datamodel/entitydef/party-entitymodel.xml
+++ b/applications/datamodel/entitydef/party-entitymodel.xml
@@ -2783,7 +2783,8 @@ under the License.
             title="Party Type Attribute">
       <field name="partyTypeId" type="id"></field>
       <field name="attrName" type="id-long"></field>
-      <field name="description" type="description"></field>
+      <field name="attrValue" type="value"></field>
+      <field name="attrDescription" type="description"></field>
       <prim-key field="partyTypeId"/>
       <prim-key field="attrName"/>
       <relation type="one" fk-name="PARTY_TYP_ATTR" rel-entity-name="PartyType">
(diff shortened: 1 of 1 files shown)


JacquesLeRoux:
Hi @stschikin ,

Please provide a description following our description template. 2 reasons here: 

1. What's the problem or rather here reason of improvement
2. Our template allows much things, like easily find the reason in Git log and fix if a problem raise later

stschikin:
Hi @JacquesLeRoux 

thanks for your advice. I have edited the original post as requested.

JacquesLeRoux:
Hi @stschikin,

Why did you remove?
```
      <relation type="many" rel-entity-name="Party">
        <key-map field-name="partyTypeId"/>
      </relation>
```

I see no pb with that.

JacquesLeRoux:
Ha, I think I know why now, because of
```
      <relation type="one" fk-name="PARTY_PTY_TYP" rel-entity-name="PartyType">
        <key-map field-name="partyTypeId"/>
      </relation>
```
Right?

stschikin:
The relation to Party seems kind of redundant with the relation to PartyType. I thought I had seen similar relations in other entities of that type, but after rechecking the TypeAttr entities (e.g. AgreementItemTypeAttr, ContactMechTypeAttr, etc), it looks like the deletion of the Party relation was unnecessary.

How shall I proceed here?

JacquesLeRoux:
Yes, we finally can keep it. A party has only one type, but a party type may have several attributes, hence several relations with a party.

JacquesLeRoux (changes requested):
Keep
```
   <relation type="many" rel-entity-name="Party">
        <key-map field-name="partyTypeId"/>
      </relation>
```

stschikin:
After re-evaluating the entire pull request, it looks like I have confused TypeAttr entities with Attribute entities. Attribute entities need a AttrValue, but TypeAttr do not. 

I found out in Jira OFBIZ-294 that TypeAttr serve as an extension to Type entities and therefore PartyTypeAttr was defined correctly and does not need any fixing...

JacquesLeRoux:
Thanks for the clarification @stschikin 

Please close your PR and the Jira, TIA

PR #801, opened 2024-05-18
Improved: Corrected the condition to fetch currency UOM id from shipment instead of shipment route segment

sandeep-rajpoot:
There is a typo in the condition while fetching a currency UOM id from the shipment route segment entity instead of shipment at the time of creating a shipment on the FedEx carrier.

Improved:Corrected the condition to fetch currency UOM id from shipment instead of shipment route segment
(OFBIZ-13013)

Explanation

Thanks:


vyasdevanshu (approved):
The change is correct as it was a typo in the fedexShipRequest service.

PR #815, opened 2024-06-12
Improved: Added support to allow direct view rendering in override vi…

dixitdeepak:
Improved: Added support to allow direct view rendering in override vi…ew functionality (OFBIZ-13117)

Added allow-direct-view-rendering in view-mapping tag, default value will be false. i.e by default now view is allowed to be used as OOTB overridden view functionality. In order to allow the view redirection (override) on all workflows allow-direct-view-rendering must be set to true. This feature may break some existing flow where overridden view workflow is used


dixitdeepak:
https://github.com/apache/ofbiz-framework/pull/816

PR #834, opened 2024-09-09
Fixed: Error when uploading image because of line length (OFBIZ-12639)

grozadanut:
For image files isValidFile should not checkMaxLinesLength, since that was created to check text files. Image validation is done by isValidImageFile

Thanks: Ingo Wolfmayr and Jacques Le Roux for the work so far in (OFBIZ-12639)


JacquesLeRoux:
I directly pushed a slightly different solution, closing here

PR #838, opened 2024-10-08
Improved: Slim down use of ${groovy: } (OFBIZ-13146)

nmalin:
Improved: Slim down use of ${groovy: } (OFBIZ-13146)

With the framework evolution, we used on many place the scriptlet ${groovy: } on screen rendering where something it's now unecessary.

This issue convert some of them on a other solution to slim down this usage


JacquesLeRoux:
Hi @nmalin,

Thanks for the good work (tested on Flat Grey for "---------- Last system notes ----------")

Should the file 
framework/common/src/main/groovy/org/apache/ofbiz/common/GetUiLables.groovy
not be named
framework/common/src/main/groovy/org/apache/ofbiz/common/GetUiLabels.groovy ;)

nmalin:
Thanks Jacques I'll have a look to many change on many subject :D

PR #850, opened 2024-11-06
Improved: Ensure unique email when sendEmail OFBIZ-13170

nmalin:
When you run the service sendCommEventAsEmail if commEvent to send contains multiple party with the same email (through different contactMech), email will be sent to many times to the same box (depending on the smtp configuration).

To avoid this case, we will ensure that an email is unique for the sender list.

### DIFF
--- a/applications/party/src/main/java/org/apache/ofbiz/party/communication/CommunicationEventServices.java
+++ b/applications/party/src/main/java/org/apache/ofbiz/party/communication/CommunicationEventServices.java
@@ -166,7 +166,7 @@ public class CommunicationEventServices {
                 }
 
                 // add other parties from roles, collect all email on map to parse it after
-                List<String> emailAlreadyLoad = UtilMisc.toList(sendTo);
+                List<String> alreadyLoaded = UtilMisc.toList(sendTo);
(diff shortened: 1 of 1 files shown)


JacquesLeRoux:
Did not test (not easy) but after simple review sounds good to me because of the use of `emailAlreadyLoad`

applications/party/src/main/java/org/apache/ofbiz/party/communication/CommunicationEventServices.java
```diff
@@ -165,45 +165,35 @@ public static Map<String, Object> sendCommEventAsEmail(DispatchContext ctx, Map<
                     return ServiceUtil.returnError(errMsg + " " + communicationEventId);
                 }
 
-                // add other parties from roles
-                String sendCc = null;
-                String sendBcc = null;
+                // add other parties from roles, collect all email on map to parse it after
+                List<String> emailAlreadyLoad = UtilMisc.toList(sendTo);
```
gtchaboussie: alreadyLoaded instead of alreadyLoad ?

nmalin:
Thanks both for remarks
@JacquesLeRoux yes it's hard to test however we have some integration test that cover any regression :)

JacquesLeRoux:
👍

PR #855, opened 2024-11-13
Fixed: Remove duplicate event handler for collapsible in modals

florianMo:
Fixed: Remove duplicate event handler for collapsible in modals

OF-13169

When a modal opens, we search for interactive elements
and trigger jQuery plugins/behaviors in bindObservers
function. This function searches for HTML elements and
starts needed behavior, using this logic:

$(wrappingElement).find(selector).each([behavior]);

This means that all 'selector' in 'wrappingElement'
initially present in the page will be triggered.
When we open a modal, we call bindObservers again
within the modal, to trigger interactive elements
which are in the modal.

For collapsible (and some others behaviors), we use

$(wrappingElement).on('click', doStuff);

...but the jQuery 'on' method also affects
DOM elements that are dynamically added to the
page, like modals content. So for these behaviors,
event handlers for interactive elements are created
twice. So when we click on the "+" of a collapsible
in a modal, "toggleCollapsiblePanel" is called twice,
and the collapsible section is open and closed right
away

### DIFF
--- a/themes/common-theme/webapp/common-theme/js/util/OfbizUtil.js
+++ b/themes/common-theme/webapp/common-theme/js/util/OfbizUtil.js
@@ -27,16 +27,16 @@ var AJAX_REQUEST_TIMEOUT = 5000;
 var SP_CLICK_ON = 'clickOn';
 
 // Add observers on DOM ready.
-$(document).ready(function() {
+$(document).ready(function () {
     // add CSRF token to jQuery AJAX calls to the same domain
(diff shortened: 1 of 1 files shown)


JacquesLeRoux:
Thanks Florian,

I tested but did not review seriously, hopefully there will be no regressions. 
If ever we will handle them. We use CTR https://www.apache.org/foundation/glossary#CommitThenReview)

florianMo:
OK @JacquesLeRoux ! Thanks

PR #867, opened 2024-12-16
Improved: Add groovydsl method : failure to return a map (OFBIZ-13195)

nmalin:
With groovy dsl we have possibility to return in one line when failure appears but we can only sent a failure message like this :

****
    return failure('bad state')
****

If we need to return information on out we need to realize it like :

****
    Map result = failure('bad state')
    result.statusId = 'BAD_STATUS'
    return result
****

Like success() we improved it to do in one line:

****
    return failure('bad state', [statusId: 'BAD_STATUS'])
****

By the way we align success function to refactoring them with default value so remove two unnecessary


JacquesLeRoux:
Hi Nicolas,

Only quickly reviewed, sounds good to me.

nmalin:
Thanks for your look !

PR #873, opened 2025-01-22
Improved: [SECURITY] (CVE-2024-36104) Path traversal leading to RCE (OFBIZ-13092)

nmalin:
Improved the denied token resolution through regexp pattern. We define each potential token
 and we generate the following regexp for each

****
  .*(%.{2,5}|[^\\w])" + token + "[^\\w].*"
****

We also improved the allowed token with analysed it form security.properties and web.xml
 directly by plain text of with sha signature to manage each special case

### DIFF
--- a/framework/security/config/security.properties
+++ b/framework/security/config/security.properties
@@ -277,13 +277,12 @@ csvformat=CSVFormat.DEFAULT
 #--
 #-- If you cross issues while loading an image file because of a token found there, you may try to surround the string by spaces, as " tr " below.
 #-- Actually most of the tokens should but it's now a bit late for me. I mean to test all of them...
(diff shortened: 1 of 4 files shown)


JacquesLeRoux:
This is for 24.09 branch only, where there is currently an issue: https://lists.apache.org/thread/8dwrx5wwkjpfm5h0xtkwtnc3rpkbyoov

JacquesLeRoux:
~~I have created https://issues.apache.org/jira/browse/OFBIZ-13204 for that.~~ It's now closed.

gilPts:
To advance, since it is no more needed, we can let aside the part about : allowed token with analysed it form security.properties and web.xml directly by plain text of with sha signature to manage each special case.

JacquesLeRoux:
Hi Guys,

Since we want to release 24.09.01, should we not close here?

JacquesLeRoux:
Seems there is no need to keep this PR open

PR #880, opened 2025-02-28
Improved: Replace printStacktrace() with Debug.logXXX (OFBIZ-13214)

mbrohl:
Removes all current occurences except those in starting classes.


nmalin:
Hello @mbrohl 
Any reason to not push your work in the code base ?

PR #883, opened 2025-03-06
Update best-bet behavior in ContentUrlTag.java

jovelromeo:
Explanation
If no.http is enable the best-bet behavior should probably be to use secure prefix.


JacquesLeRoux:
Thanks @jovelromeo,

It's OK with me. Apart for non functional changes, we a tradition of creating Jira issues. Could you please create one?
https://cwiki.apache.org/confluence/display/OFBIZ/OFBiz+Contributors+Best+Practices#OFBizContributorsBestPractices-HowtocreateaJiraissue
TIA

JacquesLeRoux:
BTW, I did not notice that you fixed the 18.12 branch. I consider this more as a bug than an improvement. So I'll merge in trunk and backport in 18.12 and 24.09.

JacquesLeRoux:
This has been handled with https://issues.apache.org/jira/browse/OFBIZ-13217
Please check it's OK with you, TIA

PR #888, opened 2025-04-28
Implemented: Fix all codenarc violation (OFBIZ-13228)

nmalin:
Waiting branch to test the OFBIZ-13228


nmalin:
Loaded from my local to update the violation authorization only for plugins.

PR #890, opened 2025-05-16
Fixed: Autocompletion on lookup with target-parameter (OFBIZ-13236)

MkLeila:
This fix a script error that makes the autocompletion fail on lookups with target-parameter (it occurs when trying to sanitize the parameter)

### DIFF
--- a/themes/common-theme/webapp/common-theme/js/util/OfbizUtil.js
+++ b/themes/common-theme/webapp/common-theme/js/util/OfbizUtil.js
@@ -926,7 +926,7 @@ function ajaxAutoCompleter(areaCsvString, showDescription, defaultMinLength, def
                 var queryArgs = { "term": request.term };
                 if (typeof args == "object" && jQuery.isArray(args)) {
                     for (var i = 0; i < args.length; i++) {
-                        queryArgs["parm" + i] = jQuery(DOMPurify.sanitize(args[i]).val())
+                        queryArgs["parm" + i] = DOMPurify.sanitize(jQuery(args[i]).val())
                     }
                 }
                 jQuery.ajax({
--- a/themes/common-theme/webapp/common-theme/js/util/fieldlookup.js
+++ b/themes/common-theme/webapp/common-theme/js/util/fieldlookup.js
@@ -304,7 +304,7 @@ var Lookup = function(options) {
         var queryArgs = "presentation=" + options.presentation;
         if (typeof options.args == "object" && jQuery.isArray(options.args)) {
             for ( var i = 0; i < options.args.length; i++) {
-                queryArgs += "&parm" + i + "=" + jQuery(DOMPurify.sanitize(options.args[i])).val();
+                queryArgs += "&parm" + i + "=" + DOMPurify.sanitize(jQuery(options.args[i]).val());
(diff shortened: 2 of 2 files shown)


JacquesLeRoux:
Hi @MkLeila,

When logged in as admin, I still get that on both Win7 and Ubuntu 20.04:

When coming to the page:
image

After entering 2 chars in the lookup ("ad")
image

MkLeila:
Hello @JacquesLeRoux,
thanks for the test:  if you've got the autocompletion working, the fix has worked correctly.

Otherwise, I also saw that tooltip error and I still don't know the root cause (I quickly checked on the form but it is not clear).
It is present both in trunk and 24.09: https://demo-trunk.ofbiz.apache.org/partymgr/control/EditCommunicationEvent, https://demo-stable.ofbiz.apache.org/partymgr/control/EditCommunicationEvent
I can try to check why it does this display

JacquesLeRoux:
Hi Leïla,

The autocompletion worked, but weirdly out of the field (in both OSs)

MkLeila:
Hello @JacquesLeRoux , what i meant is that the tooltip/lookupDescription issue is an issue that was already on this form 
I am currently trying to debug it but they are not linked (as you can see this display error is already present on demo instances without the autocompletion fix).

To explain, the problem i fixed was that the autocompletion did not work at all  (you can test it with the lookup preferredContactMechId for instance at https://demo-stable.ofbiz.apache.org/marketing/control/FindContactListParties?contactListId=9000&hideExpired=Y)
Here is an example of this non working autocompletion:
image
As you can see there was no result display at all and an error was returned when executing the script

I will continue to check for the "lookupDescription" display error to see if i can find the root cause because I find it strange too

MkLeila:
@JacquesLeRoux I think I got the root cause :-)
The field partyIdFrom from above is filled with admin and it triggers the autocompletion because the lookup from contactMechIdFrom has the parameter description-field-name="partyIdFrom"

I did not know this parameter before but it seems to be displayed on purpose. As you can see in the form, there is the comment "  this is an example of how to display different types of contact mech to the user"
I think there could be an improvement to display it differently but it seems to be volunteer

JacquesLeRoux:
Hi Leïla,

When I said 
>The autocompletion worked, but weirdly out of the field (in both OSs)

It was about the autocompletion list, not the tooltip. As you can see in my 2nd image there is no border around the list. So it looks weird. Fortunately this appears only with Edge on Win7 (yes I know :), not with FF on Ubuntu. So we can forget that.

JacquesLeRoux:
About 
>I think there could be an improvement to display it differently but it seems to be volunteer

It's very old stuff of the pre-Apache era, here is the related commit comment:
>Initial OFBiz import from revision 7923 of the old svn.ofbiz.org SVN repository.

So it's unrelated to autocompletion (implemented in 2010) and could be indeed improved. Anyway, it's just a feature demo part, not a big deal ;)

To summarize : +1 for commit :)

MkLeila:
Ah ok @JacquesLeRoux . Sorry i didn't saw there was no border (i am not used to this display)
Thanks for the test and the explanation

PR #902, opened 2025-07-27
Rest enabled

smunir:
Improved:
Implemented:
Documented:
Completed:
Reverted:
Fixed:
(OFBIZ-)

Explanation

Thanks:


mbrohl:
@smunir Thanks for your contribution.
Plugins are in a separate repository so this change is not valid for the framework.

The PR also contains custom changes like ports and properties which is also invalid.
Please use your own repository for custom changes.

Thanks.

PR #910, opened 2025-09-12
OFBIZ-13269: Add support for maximum discount cap in Cart Range Percent Discount promotions

diveshdut:
Introduced a new `maxDiscountAmount` field in `ProductPromoAction` to configure a ceiling for percentage-based cart discounts. This ensures that the applied discount never exceeds the configured maximum, regardless of cart total value.

Updated `ProductPromoActionServices.groovy` to enforce this limit during promotion calculation. Also added UI support to allow users to set the maximum discount amount when configuring promotions.

Problem
Currently, when configuring a Cart Range Percent Discount in promotions, OFBiz applies the discount percentage across the cart total without any ceiling.

This creates issues for high-value carts where the discount amount may become disproportionately large.
For example, with a 15% discount on cart total, a $2000 cart would result in a $300 discount. Many businesses want the flexibility to cap such discounts at a fixed value (e.g., $100), irrespective of cart size.

Solution:

1) Extended the ProductPromoAction entity with a new field: maxDiscountAmount.
2) Enhanced the promotion calculation logic in ProductPromoActionServices.groovy to:
    2.1) Apply the percentage discount as usual.
    2.2) Validate against the configured maximum.
    2.3) Enforce the ceiling if the computed discount exceeds maxDiscountAmount.
3) Updated the promotion setup UI to allow administrators to configure this value.

Notes:
1) Applies only when a single promotion is active on the cart (no multiple promotion stacking considered).
2) Keeps business configuration simple while addressing a common retail use case.
3) Backward-compatible: existing promotions without maxDiscountAmount remain unaffected.

### DIFF
--- a/applications/datamodel/data/demo/ManufacturingDemoData.xml
+++ b/applications/datamodel/data/demo/ManufacturingDemoData.xml
@@ -61,4 +61,7 @@ under the License.
 
     <FixedAssetDepMethod depreciationCustomMethodId="STR_LINE_DEP_FORMULA" fixedAssetId="DEMO_PROD_EQUIPMT_1"/>
     <FixedAssetDepMethod depreciationCustomMethodId="DBL_DECL_DEP_FORMULA" fixedAssetId="DEMO_PROD_EQUIPMT_2"/>
+
+    <!-- Labor Cost Data -->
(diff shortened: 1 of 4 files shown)


JacquesLeRoux:
Hi @diveshdut,

Is this PR still a draft? TIA

diveshdut:
@JacquesLeRoux this PR can be merged if this looks good to you.

JacquesLeRoux:
Thanks @diveshdut,

I'll have a look as soon as I'll be available

PR #913, opened 2025-10-07
Fixed: Decimal numbers not displayed correctly in input type=number (OFBIZ-13297)

florianMo:
Fixed: Decimal numbers not displayed correctly in input type=number
(OFBIZ-13297)

### `<input type="number"/>` does not accept decimals

Now that we use `<input type="number"/>` (https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/number, https://issues.apache.org/jira/browse/OFBIZ-13183) to display numeric values in forms, it is not possible to enter decimal values.

We need to support the `step` attribute (https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input#step), and use `any` as default value, since `any` will allow any decimal value. This means adding a new `step` attribute on the `<text/>` form field, and render it as a `step` attribute on the HTML `<input type="number"/>`.
Supporting this attribute, developer will be able to define which kind of value he expects (integer, one decimal, two decimals...).

### `<input type="number"/>` does not display decimal values correctly

`<input type="number"/>` are sensitive to the decimal separator used in their `value` attribute. Depending on the browser and the [lang](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/lang) used, an input may or may not display its value in the HTML field. For example :

* Firefox 143.03
     * `<input type="number" lang="en" value="1.2"/>` -> displays `1.2`
     * `<input type="number" lang="en" value="1,2"/>` -> displays `1,2`
     * `<input type="number" lang="fr" value="1.2"/>` -> displays `1,2`
     * `<input type="number" lang="fr" value="1,2"/>` -> displays `1,2`
 * Chrome 141
     * `<input type="number" lang="en" value="1.2"/>` -> displays `1,2`
     * `<input type="number" lang="en" value="1,2"/>` -> displays nothing
     * `<input type="number" lang="fr" value="1.2"/>` -> displays `1,2`
     * `<input type="number" lang="fr" value="1,2"/>` -> displays nothing

It vould seem reasonable to always use the same locale (`en` for example) to format numeric values before inserting it in a HTML input, and let the browser decide which decimal separator to use.

Thanks: Néréide team

### DIFF
--- a/framework/base/src/main/java/org/apache/ofbiz/base/conversion/NumberConverters.java
+++ b/framework/base/src/main/java/org/apache/ofbiz/base/conversion/NumberConverters.java
@@ -32,7 +32,7 @@ import org.apache.ofbiz.base.util.StringUtil;
 public class NumberConverters implements ConverterLoader {
 
     private static Number fromString(String str, Locale locale) throws ConversionException {
-        NumberFormat nf = NumberFormat.getNumberInstance(locale);
+        NumberFormat nf = NumberFormat.getNumberInstance(Locale.getDefault());
         if (nf instanceof DecimalFormat) {
             // CHECKSTYLE_OFF: ALMOST_ALL
             ((DecimalFormat) nf).setParseBigDecimal(true);
--- a/framework/widget/dtd/widget-form.xsd
+++ b/framework/widget/dtd/widget-form.xsd
@@ -1560,6 +1560,11 @@ under the License.
                     </xs:restriction>
                 </xs:simpleType>
             </xs:attribute>
+            <xs:attribute name="step" type="xs:string">
+                <xs:annotation>
+                    <xs:documentation>Controls the "step" attribute of the input element. Ignored if the "type" attribute of the field is not "number". See https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/step</xs:documentation>
+                </xs:annotation>
+            </xs:attribute>
             <xs:attribute name="pattern" type="xs:string">
                 <xs:annotation>
                     <xs:documentation>Controls the pattern attribute of the input element. See https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/pattern</xs:documentation>
--- a/framework/widget/src/main/java/org/apache/ofbiz/widget/model/ModelFormField.java
+++ b/framework/widget/src/main/java/org/apache/ofbiz/widget/model/ModelFormField.java
@@ -366,7 +366,7 @@ public final class ModelFormField {
             if (retVal != null) {
                 // format string based on the user's locale and time zone
                 if (retVal instanceof Double || retVal instanceof Float || retVal instanceof BigDecimal) {
-                    NumberFormat nf = NumberFormat.getInstance(locale);
+                    NumberFormat nf = NumberFormat.getInstance(Locale.getDefault());
                     nf.setMaximumFractionDigits(10);
                     return nf.format(retVal);
                 } else if (retVal instanceof java.sql.Date) {
--- a/framework/widget/src/main/java/org/apache/ofbiz/widget/model/ModelFormField.java
+++ b/framework/widget/src/main/java/org/apache/ofbiz/widget/model/ModelFormField.java
@@ -5918,6 +5918,7 @@ public final class ModelFormField {
         private final SubHyperlink subHyperlink;
         private final String type;
         private final String pattern;
+        private final String step;
 
         public TextField(Element element, ModelFormField modelFormField) {
             super(element, modelFormField);
@@ -5925,6 +5926,7 @@ public final class ModelFormField {
             this.defaultValue = FlexibleStringExpander.getInstance(element.getAttribute("default-value"));
             this.mask = element.getAttribute("mask");
             this.type = element.getAttribute("type");
+            this.step = element.getAttribute("step");
             this.pattern = element.getAttribute("pattern");
             Integer maxlength = null;
             String maxlengthStr = element.getAttribute("maxlength");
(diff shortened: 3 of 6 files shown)


JacquesLeRoux:
Hi Florian,

I just tested, we have still the load and decimals value change issue at https://localhost:8443/catalog/control/EditProductPrices?productId=GZ-1000 

Despite having 
```
<html>
<body>
<!--StartFragment-->
<input type="number" name="price"
   value="18"     size="6"                 id="UpdateProductPrice_price_o_0"                     required   
   step="any"
<!--EndFragment-->
</body>
</html>
```
So it looks like a load issue. But if you try to enter 15,99 only 15 is saved. I know it's a WIP and a draft PR :) HTH

florianMo:
Hi @JacquesLeRoux , it is related to my second point (as far as I know now, not sure yet what's happening here). When entering a decimal value in an input (dot or comma), we always get a decimal string with a dot in `NumberConverters::fromString`. If your locale is `fr`, then the parse operation will remove the decimals (I guess `parse` use only commas as decimal separator if locale is `fr`). If we use `Locale.getDefault()`, then decimals are parsed correctly.

After the value is saved, we have the same process for the display : OFBiz will grab the value from DB, and stringify it using the user locale. If the value is `1.2` in DB, it will become `1,2` (comma) in the display process (`ModelFormField::getEntry` L369). So we get an `<input type="number" step="any" value="1,2"/>` (comma), and the fact that the value contains a comma will lead to display issues depending on the browser. For example Chrome won't display this value, leading the user to believe that the value has not been saved (when in fact it has been).

I was able to "fix" this by using `Locale.getDefault()` (so `en`) on parsing (`NumberConverters::fromString`) and stringifying (`ModelFormField::getEntry`, when using `NumberFormat` on line 369), but it needs to be discussed, this is more of a quickfix to understand what's happening.

To summarize, I feel like using `<input type="number"/>` requires us to always use dots when building strings that are decimal values that will be used as an input value, so we never run into browser locale issues. And when receiving a string from a number input, we should always use the default locale as well, so dots are always considered as a decimal separator.

It should be noted that in HTML, [locale can be defined on all elements, as a `lang` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/lang). OFBiz sets the user locale on the `<html/>` element, but it can be overridden on a specific element. We could provide `lang="en"` on `<input type="number" step="any" lang="en"/>` to enforce a specific behavior. I did not test thoroughly, I'm not sure that it could be a viable solution in our case.

JacquesLeRoux:
Thanks Florian, indeed we need to take decisions. Better do that openly on dev ML, the more brains the better.

JacquesLeRoux:
Hi @florianMo, I noticed the new patch is shorter than the last one, normal?

JacquesLeRoux:
Ah OK, seems still a WIP :)

florianMo:
@JacquesLeRoux the implementation of the step attribute is ready for review, the last commit about always using `en` locale when parsing/stringifying is just a quickfix, its purpose is to demonstrate what goes wrong, providing a minimal/naive solution.

JacquesLeRoux:
Thanks Florian, naïve but working :+1:

JacquesLeRoux:
@florianMo, I'm looking at the test issue

florianMo:
thanks @JacquesLeRoux , I guess this test checks that the current timezone is used to parse numbers ? And my niave solution breaks that ?

JacquesLeRoux:
Possible indeed @florianMo , I'll get back on this afternoon, not an hurry ;)

JacquesLeRoux:
The issue is at ObjectTypeTests::simpleTypeOrObjectConvertTest L128
https://github.com/apache/ofbiz-framework/blob/7dba6ac5d87783c1209e84189d1d57996b82737e/framework/base/src/test/java/org/apache/ofbiz/base/util/ObjectTypeTests.java#L128
The wanted argument is 12,345.67. At least on a French PC the value is 12 345,67. I did not test with another language.

So, in this specific case, the problems are the decimal and thousands separator https://en.wikipedia.org/wiki/Decimal_separator#Digit_grouping

I'm not sure, but since it's only a test, maybe it's easier to force separators value, say the English ones ("," and ".') and modify the calculated value to correspond to "wanted".

Adapting the "wanted" values upstream in function of local language, eg at 
https://github.com/apache/ofbiz-framework/blob/7dba6ac5d87783c1209e84189d1d57996b82737e/framework/base/src/test/java/org/apache/ofbiz/base/util/ObjectTypeTests.java#L496
https://github.com/apache/ofbiz-framework/blob/7dba6ac5d87783c1209e84189d1d57996b82737e/framework/base/src/test/java/org/apache/ofbiz/base/util/ObjectTypeTests.java#L497
would be another solution.

I did not change nor tested anything yet.

JacquesLeRoux:
Thinking about it, only the 2nd solution is serious. The 1st one is like testing 1=1 :)

JacquesLeRoux:
Thank you for your work Florian, I'm closing after fixing the test issue.

florianMo:
Thanks @JacquesLeRoux , is there a specific reason for not merging this PR, and creating a new commit instead ? By doing this we lose traceability (who did this, the original PR description where I describe the whole thing)

JacquesLeRoux:
Hi Florian, 

You are right, I should have pushed the PR just before fixing the test issue. That would have been better. Unfortunately It's now too late as, unlike with Subversion, it's complicated to change commits comments once pushed with Git.

But don't worry people interested by these information can still found it by following the Jira and information in it, notably the PR link. On the other hand alas indeed "git blame" does not help, I'm sorry.

The principal reason is that I forgot about the PR. Mostly because I crossed an OS issue while working on the test issue as I explained at bottom of the commit comment. I finally fixed it. But, in my defense, it took me all my attention that's why I mostly forgot the rest.

florianMo:
OK @JacquesLeRoux I understand ! Not a big deal indeed.

PR #918, opened 2025-11-11
Enhance jwt validation

jacopoc:
Implemented: JWT validation for tokens issued by an external authentication server.

The system now supports two token validation modes:
1) External authentication server (JWK-based): if an issuer is configured
in the "security.token.issuer" property, the token is verified using a JWK provider and
the issuer's public key used to sign the token.
2) Local HMAC verification: If no issuer is configured, the token is verified locally using an HMAC key derived from the secret key configured in the "security.token.key" (and optionally a salt). This is the legacy mode whose behavior is not affected by this change. With the default configuration, this is the method used by OFBiz for token verification.

Change access modifiers and method signatures for token validation methods to allow upcoming implementation for external JWT validation.

Thanks: Anahita Goljahani for the analysis and research about OAuth 2.0/OpenID Connect providers and for the tests with Keycloak and its deployment and configuration.

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -76,6 +76,7 @@ dependencies {
     implementation 'oro:oro:2.0.8'
     implementation 'wsdl4j:wsdl4j:1.6.3'
     implementation 'com.auth0:java-jwt:4.4.0'
+    implementation 'com.auth0:jwks-rsa:0.22.2'
     implementation 'org.jdom:jdom2:2.0.6.1'
     implementation 'com.google.re2j:re2j:1.7'
     implementation 'xerces:xercesImpl:2.12.2'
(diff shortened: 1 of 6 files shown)


jacopoc:
Just for the record, the three unit tests that are failing after my changes are doing so because they were previously passing only due to a side effect. One of the methods in JWTManager is incorrectly implemented:
```
    /**
     * Get the JWT secret key from database or security.properties.
     * @param delegator the delegator
     * @return the JWT secret key
     */
    private static String getJWTKey(Delegator delegator, String salt) {
        String key = UtilProperties.getPropertyValue("security", "security.token.key");
        if (key.length() < 64) { // The key must be 512 bits (ie 64 chars)  as we use HMAC512 to create the token, cf. OFBIZ-12724
            throw new SecurityException("The JWT secret key is too short. It must be at least 512 bites.");
        }
        if (salt != null) {
            return StringUtil.toHexString(salt.getBytes()) + key;
        }
        return key;
    }
```
As you can see, the method accepts a Delegator object and claims to read the secret key from either the database or security.properties, but in reality, it only reads from security.properties — the Delegator is not used at all.

Therefore, the three tests that simply pass a mock Delegator were not failing before because getJWTKey wasn’t actually using it. Now, however, the tests are failing because my updated code correctly fetches the settings from the database or security.properties, and therefore does require a valid Delegator object.

I’ll look into how we can improve the tests and then fix getJWTKey too.

jacopoc:
Ok, all checks are successful now after my last commit that fixes the setUp method in ModelFormTest. This pull request can now be merged.

PR #919, opened 2025-11-12
docker-entrypoint.sh not properly handling variable OFBIZ_DISABLE_COMPONENTS

javier-ochoa:
Fixed: docker-entrypoint.sh does not properly handle variable OFBIZ_DISABLE_COMPONENTS (OFBIZ-13314)

docker-entrypoint.sh did not properly handle variable OFBIZ_DISABLE_COMPONENTS, but it is fixed so that the variable can be accepted being a list of comma-separated values, and disable one after another.

Thanks: Javier Ochoa

### DIFF
--- a/docker/docker-entrypoint.sh
+++ b/docker/docker-entrypoint.sh
@@ -146,7 +146,7 @@ ofbiz_setup_env() {
   OFBIZ_POSTGRES_TENANT_USER=${OFBIZ_POSTGRES_TENANT_USER:-ofbiztenant}
   OFBIZ_POSTGRES_TENANT_PASSWORD=${OFBIZ_POSTGRES_TENANT_PASSWORD:-ofbiztenant}
 
-  OFBIZ_DISABLE_COMPONENTS=${OFBIZ_DISABLE_COMPONENTS-plugins/birt/ofbiz-component.xml}
+  OFBIZ_DISABLE_COMPONENTS=${OFBIZ_DISABLE_COMPONENTS:-plugins/birt/ofbiz-component.xml}
 }
 
 ###############################################################################
@@ -260,7 +260,6 @@ disable_component() {
   XML_FILE="/ofbiz/$1"
   if [ -f "$XML_FILE" ]; then
     TMPFILE=$(mktemp)
-
     xsltproc /ofbiz/disable-component.xslt "$XML_FILE" > "$TMPFILE"
     mv "$TMPFILE" "$XML_FILE"
   else
@@ -273,19 +272,24 @@ disable_component() {
 # components' 'enabled' attribute to false.
 # $1 - Comma separated list of paths to configuration XML files to be modified.
 disable_components() {
-  COMMA_SEPARATED_PATHS="$1"
+  COMMA_SEPARATED_PATHS=$1
 
-  if [ -n "$COMMA_SEPARATED_PATHS" ]; then
+  # Remove spaces after commas
+  COMMA_SEPARATED_PATHS="${COMMA_SEPARATED_PATHS//, /,}"
 
-    # Split the comma separated paths into separate arguments.
-    IFS=,
-    set "$COMMA_SEPARATED_PATHS"
+  [ -z "$COMMA_SEPARATED_PATHS" ] && return 0
 
-    while [ -n "$1" ]; do
-      disable_component "$1"
-      shift
-    done
-  fi
+  # Split on commas into $1 $2 ...
+  oldIFS=$IFS
+  IFS=,
+  set -- $COMMA_SEPARATED_PATHS    # <- no quotes, use -- to avoid option parsing
+  IFS=$oldIFS
+
+  # Iterate through the split arguments
+  while [ -n "$1" ]; do
+    disable_component "$1"
+    shift
+  done
 }
 
 ###############################################################################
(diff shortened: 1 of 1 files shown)


docker/docker-entrypoint.sh
```diff
@@ -146,7 +146,7 @@ ofbiz_setup_env() {
   OFBIZ_POSTGRES_TENANT_USER=${OFBIZ_POSTGRES_TENANT_USER:-ofbiztenant}
   OFBIZ_POSTGRES_TENANT_PASSWORD=${OFBIZ_POSTGRES_TENANT_PASSWORD:-ofbiztenant}
 
-  OFBIZ_DISABLE_COMPONENTS=${OFBIZ_DISABLE_COMPONENTS-plugins/birt/ofbiz-component.xml}
+  OFBIZ_DISABLE_COMPONENTS=${OFBIZ_DISABLE_COMPONENTS:-plugins/birt/ofbiz-component.xml}
```
danwatford, line 149: This means it will not be possible to prevent any components from being disabled by setting OFBIZ_DISABLE_COMPONENTS=.

Did you mean to always require at least one component to always be disabled?

javier-ochoa: hello @danwatford. Perhaps I made a mistake since sh scripts are not my strength, but i wonder what is the difference. As far as I know, the original expression would use default value only if the variable was not set. If the variable was set, and empty, then it would use empty, which can be argued if it is ok or not. Adding double-dots, if I am not wrong, the variable must be set and be different than empty, or else, the default value would be used. 

So the diff is if the default value should be used when variable is set but empty (OFBIZ_DISABLE_COMPONENTS="")

Nevertheless, if it should be without the dots, personally i dont have any issue with the rollback of the change.

Cheers!

danwatford: Hi @javier-ochoa - 

> As far as I know, the original expression would use default value only if the variable was not set. If the variable was set, and empty, then it would use empty, which can be argued if it is ok or not.

With the change in this PR, how do we handle the case where the user wants to enable the birt plugin?

JacquesLeRoux:
Hi guys,

Indeed it's annoying to not be able to enable only Birt when using Docker. Though we must know what that means.

Long story short, the Birt component was disabled with [OFBIZ-12824](https://issues.apache.org/jira/browse/OFBIZ-12824) because of  CVE-2022-25371. That does not prevent to use Birt in OFBiz. But so as far as we have not done the work needed by CVE-2014-3596 (see [OFBIZ-5744](https://issues.apache.org/jira/browse/OFBIZ-5744) for details) it can be used only locally (without any Internet connection) without fearing security attacks.

Apart some rare cases*, the Birt component  is mostly used for [flexible report](https://github.com/apache/ofbiz-plugins/blob/f8121753152a9317c302731e346475958984ffee/birt/src/docs/asciidoc/birt.adoc#L4). 

\* [If you search for the word Birt in Jira](https://issues.apache.org/jira/browse/OFBIZ-12231?jql=project%20%3D%20OFBIZ%20AND%20status%20in%20(Open%2C%20%22In%20Progress%22%2C%20Reopened%2C%20%22Patch%20Available%22)%20AND%20text%20~%20%22birt%22)
Only
OFBIZ-9340
OFBIZ-9382
OFBIZ-11079
OFBIZ-12231
are related

To allow OFBiz users to use the birt component when using Docker, I suggest to comment in docker-entrypoint.sh about OFBIZ-13314 to allow them to revert the https://github.com/apache/ofbiz-framework/commit/c81c469ec4 commot. It's not even a big deal to do it by hand.

What do you think ?

JacquesLeRoux:
Hi, I'll do that

JacquesLeRoux:
This is the comment I propose:
```
###############################################################################
# With https://issues.apache.org/jira/browse/OFBIZ-13314 a change has been here to allow disabling several components
# A side effect is that you need to disable at least one component. By defaut it's birt component.
# It's then complicate if you want to enable the birt component without disabling any other component.
# A solution for that is to revert the changes done by https://github.com/apache/ofbiz-framework/pull/919
# Fortunately it's also easy to be done by hand.
  OFBIZ_DISABLE_COMPONENTS=${OFBIZ_DISABLE_COMPONENTS:-plugins/birt/ofbiz-component.xml}
}
```
Please confirm it's OK with you

JacquesLeRoux:
Pushed with https://github.com/apache/ofbiz-framework/commit/ef90449a96145691b0ba9e6f61b3435487118906

danwatford:
Hi @JacquesLeRoux  -  I have been trying to find the time to answer on this topic for days, but unfortunately didn't get to you in time.

I don't think you should have the comment. Instead we should just revert that line back to the original value so that it is possible to leave all components enabled.

Asking users of the docker container to change the entrypoint.sh script means they need to understand how the script works, and possibly rebuild the container themselves. This means they won't be able to use the pre-built containers provided by the project as a convenience.

JacquesLeRoux:
Hi Dan,

Thanks for your feedback. As I have not used the script, do you mean that's only this line needs to be reverted? Will that allow to disable several components?

danwatford:
> Hi Dan,
> 
> Thanks for your feedback. As I have not used the script, do you mean that's only this line needs to be reverted? Will that allow to disable several components?

Yes - only the line in the comment at the following link needs to be reverted:
https://github.com/apache/ofbiz-framework/pull/919#discussion_r2529812823

JacquesLeRoux:
Thanks Dan,

As you wrote this script I trust you are the best person to decide that :)
I'll revert my comment and the assigning OFBIZ_DISABLE_COMPONENTS

PR #923, opened 2025-11-24
OFBIZ-13313

gtchaboussie:
EntityQuery use dispatcher

Improved: Creates an interface for queryable objects, or objects that have a delegator.
(OFBIZ-13313)

Explanation:  Allows the use of the dispatcher directly in the EntityQuery.
Created the Queryable interface to do so.

Thanks: to Jacopo and Nicolas for the feedback.


framework/entity/src/main/java/org/apache/ofbiz/entity/Delegator.java
```diff
@@ -43,15 +43,18 @@
...
 import org.w3c.dom.Document;
 import org.w3c.dom.Element;
 import org.xml.sax.SAXException;
 
-public interface Delegator {
+public interface Delegator extends Queryable {
 
     enum OperationType { INSERT, UPDATE, DELETE }
 
+    Delegator getDelegator();
```
jacopoc: This declaration is not required since it is an inherited method.

framework/entity/src/main/java/org/apache/ofbiz/entity/util/Queryable.java
```diff
@@ -0,0 +1,17 @@
+package org.apache.ofbiz.entity.util;
+
+import org.apache.ofbiz.entity.Delegator;
+
+/**
+ * Interface used for the <code>EntityQuery</code> to set the delegator to yse for the query.
```
jacopoc: The Javadoc should just describe the capability of the interface without directly mentioning the classes that may use it (such as EntityQuery). There are also a few typos, such as: "yse", "easyer".

framework/entity/src/main/java/org/apache/ofbiz/entity/util/Queryable.java
```diff
@@ -0,0 +1,17 @@
...
+ * Allows easyer and shorter query instructions, like calling an <code>EntityQuery</code> with a dispatcher.
+ */
+public interface Queryable {
+
+    /**
+     * Gets the GenericEntityDelegator associated with this dispatcher
```
jacopoc: The Javadoc is not correct as it mentions GenericEntityDelegator rather than Delegator and it mentions "this dispatcher" which is out of context in this interface.

jacopoc:
Thanks Gaetan,

The introduction of the new interface makes sense and solves the concern I had.
However in my opinion the name "Queryable" doesn't describe very well the capability of the new interface. A better name could be "DelegatorProvider".
I have also added a few comments inline.

gtchaboussie:
Thanks for the review, all issues should have been fixed.
The name of "DelegatorProvider" sounds fine to me, and more descriptive of the intention of the interface.

jacopoc:
Thanks Gaetan, it looks good to me now.

nmalin (changes requested):
Hello Gaëtan,

I added some improvement on the following patch.
* Add Licence Apache 2.0 on header for DelegatorProvider
* GenericEntity implements DelegatorProvider
* Add test on EntityQuery to be sure that the DelegatorProvider and EntityQuery.use(DelegatorFactory) work well

Thanks for your work !

[OFBiz-PR923.patch](https://github.com/user-attachments/files/24879276/OFBiz-PR923.patch)

PR #924, opened 2025-11-26
OFBIZ-13070 - Fixed missing custom period start by fixing the conditi…

pavololbert:
Fixed: BalanceSheet generates an error (OFBIZ-13070)

Fixes missing custom period start by fixing the condition for fromData

Due to the change from the previous version 22.01, there was a change in the PeriodServices and the condition for finding the period was wrong causing the exception described in the bug.

The correct implementation of the condition was taken from:
https://github.com/apache/ofbiz-framework/blob/release22.01/applications/accounting/minilang/period/PeriodServices.xml#L23

Thanks: https://github.com/pavololbert

### DIFF
--- a/applications/accounting/src/main/groovy/org/apache/ofbiz/accounting/period/GPeriodServices.groovy
+++ b/applications/accounting/src/main/groovy/org/apache/ofbiz/accounting/period/GPeriodServices.groovy
@@ -33,7 +33,7 @@ Map findCustomTimePeriods() {
         List parentOrganizationPartyIdList = serviceResult.parentOrganizationPartyIdList
         EntityCondition condition = new EntityConditionBuilder().AND {
             IN(organizationPartyId: parentOrganizationPartyIdList)
-            LESS_THAN(fromDate: parameters.findDate)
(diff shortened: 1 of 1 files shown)


JacquesLeRoux:
This is no related to https://issues.apache.org/jira/browse/OFBIZ-13070

pavololbert:
> This is no related to https://issues.apache.org/jira/browse/OFBIZ-13070

This is related to this problem from my point of view, because the balance functionality underhood calls this method to get the cutom period and since this wrong condition, nothing is returned, see the stack trace and message
The following required parameter is missing: customTimePeriodStartDate

JacquesLeRoux:
OK thanks, I'll check that. BTW does it fix the issue reported by Pierre?

pavololbert:
Yes, both reports (BalanceSheet and ComparativeBalanceSheet) mentioned in the bug are fixed by this :) the issue is present in trunk but also 24.09 release

JacquesLeRoux:
Great, thanks!

JacquesLeRoux:
Checked  it's OK, got other unrelated issues, will create Jiras for them

PR #928, opened 2025-11-29
Ofbiz-13322 removed useless collections

dk2k:
Improved:
Implemented:
Documented:
Completed:
Reverted:
Fixed:
(OFBIZ-13322)

Explanation

Thanks:

### DIFF
--- a/applications/content/src/main/java/org/apache/ofbiz/content/content/ContentSearch.java
+++ b/applications/content/src/main/java/org/apache/ofbiz/content/content/ContentSearch.java
@@ -118,7 +118,6 @@ public class ContentSearch {
     public static class ContentSearchContext {
         private int index = 1;
         private List<EntityCondition> entityConditionList = new LinkedList<>();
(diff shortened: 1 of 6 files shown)


JacquesLeRoux:
Hi @dk2k,

I agree that all these removing make sense. But I prefer to keep them even if they are currently indeed useless. I can't believe that these codes have been added without reason. They are all pre-Apache era. Maybe some have been incorrectly modified before. And maybe at some point someone will detect that. That's why I prefer to keep, they don't hurt and could maybe useful in future, even if very improbably.

If you disagree, please start a thread on OFBiz dev ML dev@ofbiz.apache.org. Else please close this PR.

Thanks for all your work so far.

dk2k:
@JacquesLeRoux I understand. I will keep this PR open for some time. And even if I close it, you can use it later as a bug report.

JacquesLeRoux:
OK, thanks! 

I have closed the Jira as "Pending Closed" even if I'm not sure what's the diff with Closed and what it means :)

PR #934, opened 2025-12-29
Fixed: Force Complete Purchase Order logic to support items split into multiple ship groups (OFBIZ-13327)

golja:
The fix is needed because the cancelOrderItem service, that is used in the 'Force Complete Purchase Order' workflow, doesn't work correctly in case of item's units split into multiple ship groups.

The issue can be reproduced as described in the attached document.

[PR_force_complete_order_how_to_reproduce_the_issue.pdf](https://github.com/user-attachments/files/24367999/PR_force_complete_order_how_to_reproduce_the_issue.pdf)

### DIFF
--- a/applications/order/src/main/java/org/apache/ofbiz/order/order/OrderServices.java
+++ b/applications/order/src/main/java/org/apache/ofbiz/order/order/OrderServices.java
@@ -2163,6 +2163,12 @@ public class OrderServices {
 
         if (orderItemShipGroupAssocs != null) {
             for (GenericValue orderItemShipGroupAssoc : orderItemShipGroupAssocs) {
+                BigDecimal aisgaCancelQuantity = orderItemShipGroupAssoc.getBigDecimal("cancelQuantity");
+                if (aisgaCancelQuantity == null) {
+                    aisgaCancelQuantity = BigDecimal.ZERO;
+                }
+                BigDecimal availableQuantity = orderItemShipGroupAssoc.getBigDecimal("quantity").subtract(aisgaCancelQuantity);
+
                 GenericValue orderItem = null;
                 String itemStatus = "ITEM_CANCELLED";
                 try {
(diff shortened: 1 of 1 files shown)


JacquesLeRoux:
Hi @golja,

After patching, creating 2 ship groups and moving a part (10 items) in the 2nd group, I pushed the force button and checked the order. All items are cancelled. Is that expected?

It's the same receiving or not before pushing the force button.

golja:
Hi Jacques (happy new year!),

yes, I can confirm that canceling all of the order’s 'Remaining' items when clicking 'Force Complete Purchase Order' is the intended behavior. When an order is canceled, the expectation is that all remaining items in the order are canceled, not only those belonging to a specific Ship Group.

By the way, this is the same behavior that existed before my changes for managing items split across multiple Ship Groups. You can verify this in the trunk/demo by creating an order with two items, each assigned to a different Ship Group, and, then, using  the 'Force Complete Purchase Order' button before and after receiving items. 

In my opinion, the potentially misleading aspect of the' Force Complete Purchase Order' button is its placement: it should appear only on the Purchase Order details screen, and not also on the Receive Against PO screen. I am planning to submit additional PRs, after this one is merged, in order to address this aspect along with the other issues.

Please let me know if you need any further clarification.

Anahita

JacquesLeRoux:
Thanks Anahita, 

I got it, just that  cancelling all items when pushing 'Force Complete Purchase Order' does not make sense to me. Better name it  'Cancel all items', or do I miss something?

golja:
I think that the label 'Force Complete Purchase Order' is acceptable since the completion of the order is 'forced' by cancelling all Remaining (not yet received) items units. I would rather remove the 'Force Complete Purchase Order' button from the Receive Against PO screen, because the action of forcing the completion of the order impacts also the other Ship Groups of the order.

JacquesLeRoux:
Thanks Anahita,

OK, pushing

PR #936, opened 2026-01-13
Fixed: Form fields are not properly populated with values from the context due to a regression introduced by commit 69697d1 (OFBiz-13331)

golja:
This commit restores the condition that determines whether request or context parameters are used to the behavior prior to commit 69697d1. It also implements the new useRequestParameters.<form_name> context attribute for ModelGrid forms, which was introduced in that commit.

### DIFF
--- a/framework/widget/src/main/java/org/apache/ofbiz/widget/model/ModelFormField.java
+++ b/framework/widget/src/main/java/org/apache/ofbiz/widget/model/ModelFormField.java
@@ -289,8 +289,10 @@ public final class ModelFormField {
     }
 
     public String getEntry(Map<String, ? extends Object> context, String defaultValue) {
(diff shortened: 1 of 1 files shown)


jacopoc:
I have verified this pull request: it looks good to me and it fixes a relevant regression.
In fact, since commit 69697d1, all the forms of OFBiz trunk are broken as you can see from the screenshot below in which the form fields are not populated with the data from the ProductStore record.

Considering the severity of the issue, I suggest to merge this pull request as soon as possible. I could do it but I'd like to wait for @nmalin to verify if the use case addressed by his commit 69697d1 and described in OFBIZ-13331 is properly supported by this pull request.

Screenshot 2026-01-13 at 22 46 22

jacopoc:
@nmalin I decided to merge the PR to restore the form functionality, which was currently broken on trunk and on our demo instance. You can now review your work on OFBIZ-13331 without any time pressure.

PR #946, opened 2026-01-22
Improved: Avoid adding functions on jQuery object, remove use of internal jQuery UI function (OFBIZ-13346)

florianMo:
- Rewrite getForm and getFormField functions as standalone functions instead of adding them on
the jQuery prototype
- Update usages
- Remove usage of jQuery UI "form" internal method

(OFBIZ-13346)

Thanks: Néréide team


gtchaboussie:
Looks ok to me

PR #949, opened 2026-01-29
FIX: port offset applied twice when loading WebSite properties (OFBIZ-13352)

Lukas-Finster:
Fixed: port offset applied twice when loading WebSite properties (OFBIZ-13352)

Explanation:
In the method from(GenericValue), web site default settings are loaded. Configurations are overridden by the WebSite being used. In particular, a port offset is always added here, which can be set via the start parameters. Therefore, it is not necessary to adjust the WebSite entry in the database to store an alternative port to 8443. The method from(HttpServletRequest) calls the method from(GenericValue) itself and additionally sets the port offset. The result is that an offset is added twice.

Now addPortoffset is initialized with false and only set to true when new WebSiteProperties are created.

### DIFF
--- a/framework/webapp/src/main/java/org/apache/ofbiz/webapp/website/WebSiteProperties.java
+++ b/framework/webapp/src/main/java/org/apache/ofbiz/webapp/website/WebSiteProperties.java
@@ -81,7 +81,7 @@ public final class WebSiteProperties {
         Assert.notNull("request", request);
         WebSiteProperties webSiteProps = (WebSiteProperties) request.getAttribute("_WEBSITE_PROPS_");
         if (webSiteProps == null) {
-            Boolean addPortoffset = true;
+            Boolean addPortoffset = false;
             Delegator delegator = (Delegator) request.getAttribute("delegator");
             if (delegator != null) {
                 String webSiteId = WebSiteWorker.getWebSiteId(request);
@@ -94,6 +94,7 @@ public final class WebSiteProperties {
             }
             if (webSiteProps == null) {
                 webSiteProps = new WebSiteProperties(delegator);
+                addPortoffset = true;
             }
             if (webSiteProps.getHttpPort().isEmpty() && !request.isSecure()) {
                 webSiteProps = webSiteProps.updateHttpPort(String.valueOf(request.getServerPort()));
@@ -103,7 +104,6 @@ public final class WebSiteProperties {
             }
             if (webSiteProps.getHttpsPort().isEmpty() && request.isSecure()) {
                 webSiteProps = webSiteProps.updateHttpsPort(String.valueOf(request.getServerPort()));
-                addPortoffset = false; // We take the port from the request, don't add the portOffset
             }
             if (webSiteProps.getHttpsHost().isEmpty()) {
                 webSiteProps = webSiteProps.updateHttpsHost(request.getServerName());


toaditi:
Thanks for picking this up, @Lukas-Finster — I tested the PR locally and it does fix the HTTPS case nicely 👍

I ran OFBiz with `--portoffset=10000` (so http→18080, https→18443) with a `WebSite` entity present for the ecommerce webapp (`webSiteId=WebStore`), and logged the ports that `WebSiteProperties.from(request)` actually computes:

| | httpPort | httpsPort |
|---|---|---|
| trunk (before) | 28080 | 28443 |
| this PR | 28080 | **18443** ✅ |

So the secure redirect now correctly lands on `https://localhost:18443/…` instead of the dead `:28443` — the main symptom is fixed.

One small thing I noticed while testing: the **http** port still gets offset twice (28080). It seems to be because `addPortOffset(boolean addHttpsOffset)` always offsets the http port (the flag only guards the https part), and `from(request)` still calls it on the WebSite-entity path where `from(GenericValue)` has already applied the offset. So full-path *non-secure* URLs (order-confirmation emails, sitemap/RSS, etc.) would still be off by one offset.

If it's useful, one way to cover both is to skip that second call entirely on the WebSite path, and keep a separate flag just for the https-from-request case:

```java
boolean applyPortOffset = false;   // from(GenericValue) already offsets; only offset here on the url.properties fallback
boolean addHttpsOffset = true;
...
if (webSiteProps == null) {
    webSiteProps = new WebSiteProperties(delegator);
    applyPortOffset = true;
}
...
if (webSiteProps.getHttpsPort().isEmpty() && request.isSecure()) {
    webSiteProps = webSiteProps.updateHttpsPort(String.valueOf(request.getServerPort()));
    addHttpsOffset = false; // port already comes from the request
}
...
if (applyPortOffset) {
    webSiteProps = webSiteProps.addPortOffset(addHttpsOffset);
}
```

With that, both ports come out single-offset (18080 / 18443) in my test. I put it together in #1392 in case it's helpful to compare — but it's totally your call which direction to take, and thanks again for taking this on.

(One note for anyone reproducing: a `WebSite` entity has to exist for the affected `webSiteId` — with a seed-only DB everything falls back to the default path and the bug doesn't show.)

arunpati:
We can close this PR as changes has been covered in PR- https://github.com/apache/ofbiz-framework/pull/1392

Closing this PR. Thanks @Lukas-Finster @toaditi  for your contribution

jacopoc:
@arunpati 

I'd like to point out something about the process, as I think it's important for the health of the project and the contributor experience.

In this case, the original implementation was authored by @Lukas-Finster and contributed in this PR. A follow-up PR was then opened by @toaditi with additional changes based on the review, but it didn't acknowledge the original author's work in the commit message. The original PR was subsequently closed, and the follow-up PR was merged.

While the final code is what matters, this workflow can unintentionally deprive the original contributor of recognition for their work. A more contributor-friendly approach would have been to either:

* (preferred) continue working on the original PR (this one) by requesting the additional changes there, or
* if a replacement PR was necessary, clearly credit the original author in the commit message, PR description, and/or merge commit.

Giving proper attribution encourages contributions and helps maintain a welcoming and collaborative open-source community.

I hope we can keep this in mind for similar situations in the future.

arunpati:
Thanks @jacopoc  for raising this. I agree that giving proper credit to contributors is important for maintaining a healthy and welcoming open-source community.

I agree that it would have been better to either continue with the original PR or explicitly acknowledge the original author's contribution in the PR and commit message.

Thanks for bringing this up in a constructive way. I will keep this in mind to ensure contributors receive appropriate recognition for their work.

@toaditi, let's keep this in mind for future contributions as well.

@Lukas-Finster , Thank you for your valuable contribution to the project, and sorry that the attribution could have been better.

Lukas-Finster:
@arunpati, @toaditi  No worries, small mistakes are a part of life :)

PR #953, opened 2026-01-30
Improved: Lookup don't work on area just updated (OFBIZ-13332)

MkLeila:
Improved: Lookup don't work on area just updated (OFBIZ-13332)

fix to avoid bindObserver events duplication when calling updateArea


jacopoc:
@MkLeila, since this pull request contains a merge commit (0f2add1589ebf2cdb8c9df7cc94b5fa93f557fc6), please use "Squash and merge" rather than "Rebase and merge" when integrating it into trunk, in order to avoid introducing spurious commits.

PR #963, opened 2026-02-13
Fixed: Ensure that the invoice item quantity is set to 1 only if the quantity is null when invoice items are created by the createInvoiceItem service (OFBIZ-13362)

golja:
This PR solves the first issue described in [OFBIZ-13362](https://issues.apache.org/jira/browse/OFBIZ-13362)

When the system creates a receipt with an accepted quantity equal to zero, the service createInvoiceItem erroneously set the item invoice quantity to 1, leading to a wrong invoiced quantity.

The change in the PR makes sure that only quantities equal to null are set to 1.


jacopoc:
@golja it would be great if you could provide a PR to backport this fix to release24.09 (in preparation for the upcoming release 24.09.06).

golja:
Done! 
PR #1073

PR #964, opened 2026-02-13
Fixed: Avoid displaying quantities equal to zero as 1 in the invoice items grid of the View Invoice screen (OFBIZ-13362)

golja:
This PR solves the second issue described in [OFBIZ-13362](https://issues.apache.org/jira/browse/OFBIZ-13362)

Even when the invoice item quantity is correctly set to zero in the database, the screen viewInvoice erroneously ‘visualize’ an invoice item reporting one unit of the product.

The problem is due to the fact that the grid 'ListInvoiceItems' used in the viewInvoice screen set the invoice item quantity to one when it is zero. The change in the PR makes sure that the quantity visualized is set to 1 only when the quantity is null.


jacopoc:
@golja it would be great if you could provide a PR to backport this fix to release24.09 (in preparation for the upcoming release 24.09.06).

golja:
Done! 
PR #1073

PR #968, opened 2026-02-25
(OFBIZ-13366) : Improved performance and implementation of database-level pagination in FindGeneric.groovy

pushprajsinghjadoun:
Improved: Improved performance and implementation of database-level pagination in FindGeneric.groovy (OFBIZ-13366)

Efficiency of the generic entity search and listing functionality within WebTools (FindGeneric.groovy) when handling large datasets.

Implemented:
Database-level pagination using limit() and offset() via direct EntityQuery execution in the Groovy script, eliminating in-memory filtering for large result sets.

Completed:
Refactoring of the data retrieval logic to ensure EntityListIterator is used with a fixed result size, and updating the dynamic form generation to support pre-paginated data.

Fixed:
Performance-related system timeouts and memory exhaustion issues encountered when performing searches on large entities (e.g., Product) without search conditions.

Explanation

The previous implementation of FindGeneric.groovy relied on the performFind service within the dynamically generated XML form’s <actions> block. This approach was inefficient for large datasets because pagination was not consistently applied at the database level, often leading to excessive memory usage, timeouts, or server crashes when browsing large entities without filters.

This change moves the data retrieval logic directly into the Groovy script. By using EntityQuery with explicit limit and offset values derived from the VIEW_INDEX and VIEW_SIZE request parameters, only the required subset of records is fetched from the database. Additionally, getResultsSizeAfterPartialList() is used to provide accurate total counts for the UI without loading all records into memory, and the dynamic form renderer is informed that the data is already paginated via the override-list-size attribute. Support for suffixed pagination parameters (e.g., VIEW_INDEX_1) has also been added.

Thanks: The OFBiz community and reviewers for their feedback and guidance.

### DIFF
--- a/framework/webtools/src/main/groovy/org/apache/ofbiz/webtools/entity/FindGeneric.groovy
+++ b/framework/webtools/src/main/groovy/org/apache/ofbiz/webtools/entity/FindGeneric.groovy
@@ -121,6 +121,8 @@ if (modelEntity) {
 
         if (prepareFindResult.orderByList) {
             query.orderBy(prepareFindResult.orderByList)
+        } else {
+            query.orderBy(modelEntity.getPkFieldNames())
         }
 
         if (fieldsToSelect) {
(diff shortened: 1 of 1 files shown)


nmalin:
I merged and pushed it from my local.
Thanks for this PR @pushprajsinghjadoun

PR #969, opened 2026-02-28
Implemented: Remove demo secret keys from security.property and add a gradle task to generate and set the keys

jacopoc:
The new task, generateSecretKeys, is automatically triggered when the loadAll is executed.


JacquesLeRoux:
Hi @jacopoc,

I vaguely remember I suggested some years ago to use a secret key generator rather than stocking it in a properties file.
I reviewed your PR and it seems quite good to me. 
Just a question, why not a related Jira?

golja:
CVE-2026-31986

PR #970, opened 2026-03-05
Fixed : Refactor getMultiPartParameterMap to lazy getter to prevent re-parsing consumed input stream (OFBIZ-13369)

Lukas-Finster:
Fixed:  Refactor getMultiPartParameterMap to lazy getter to prevent re-parsing consumed input stream
(OFBIZ-13369)

Explanation: With [OFBIZ-13295](https://issues.apache.org/jira/browse/OFBIZ-13295) ControlFilter#doFilter called UtilHttp#getParameterMap, consuming the request's input stream in the process. Since a multipart stream can only be read once, all subsequent calls return an empty map, which lead to a problem with file uploads in our local setup.

UtilHttp#getMultiPartParameterMap was refactored into a lazy getter: It first checks whether the result is already cached as the request attribute "multiPartMap", and only parses the stream on the first call

Thanks:

### DIFF
--- a/framework/base/src/main/java/org/apache/ofbiz/base/util/UtilHttp.java
+++ b/framework/base/src/main/java/org/apache/ofbiz/base/util/UtilHttp.java
@@ -172,9 +172,6 @@ public final class UtilHttp {
             throw new RuntimeException(e);
         }
         params.putAll(multiPartMap);
-        if (req.getAttribute("multiPartMap") == null) {
-            req.setAttribute("multiPartMap", multiPartMap);
-        }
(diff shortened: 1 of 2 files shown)


nmalin:
Hello @Lukas-Finster, Can I reproduce the problem on demo that you fix solve ?

Lukas-Finster:
Hello @nmalin,
currently there is not a known bug that this fixes. We noticed the described in our custom setup and wanted to provide the improvement preemptivly for the standard. Describing it as an improvement instead of a fix would have been more accurate.

arunpati (approved):

PR #981, opened 2026-03-12
Align securedLoginId cookie processing with JWT-based verification

ashishvijaywargiya:
Improved: Align securedLoginId cookie processing with JWT-based verification


golja:
CVE-2026-31387

PR #983, opened 2026-03-13
Improved: Allow opting-out from the whitelist-restricted static models in freemarker (OFBIZ-13371)

Criux:
Improved: Allow the use of the whitelist restricted static Models in Freemarker to be globally switched off through the property freemarker.use-restricted-static-models in security.properties.

[(OFBIZ-13371)](https://issues.apache.org/jira/browse/OFBIZ-13371)

Explanation: Currently in trunk and 24.09.06 all static method (shared by the "Static" variable) in Freemarker templates need to be explicitly whitelisted so they can be called.

Although this gives a lot of control over the execution context inside the freemarker templates, it might translate to a lot of effort for existing projects that rely heavily on custom plugins to upgrade to this version without introducing breaking changes, since the maintainers or plugin authors would have to create a comprehensive list of all the fully qualified names of methods used so that it can be included in the whitelist.

For this reason, I propose to allow the whitelist to be turned off  globally through a security property (enabled by default), to allow all projects to benefit from upgrading to this version but also be able to plan the effort for updating their whitelists at a more comfortable pace.

In any case, the developers that decide to go back to the unrestricted "Static" variable, should be warned in a comment about the implications of their decision.

### DIFF
--- a/framework/base/src/main/java/org/apache/ofbiz/base/util/template/FreeMarkerWorker.java
+++ b/framework/base/src/main/java/org/apache/ofbiz/base/util/template/FreeMarkerWorker.java
@@ -82,8 +82,8 @@ public final class FreeMarkerWorker {
     private static final UtilCache<String, Template> CACHED_TEMPLATES =
             UtilCache.createUtilCache("template.ftl.general", 0, 0, false);
     private static final BeansWrapper DEFAULT_OFBIZ_WRAPPER = new BeansWrapperBuilder(VERSION).build();
-    private static final TemplateHashModel DEFAULT_RESTRICTED_STATIC_MODELS =
-            RestrictedStaticModels.fromConfig(DEFAULT_OFBIZ_WRAPPER.getStaticModels(), "freemarker-whitelist");
+    private static final TemplateHashModel DEFAULT_STATIC_MODELS =
+            getConfiguredStaticModel(getDefaultOfbizWrapper());
     private static final Configuration DEFAULT_OFBIZ_CONFIG = makeConfiguration(DEFAULT_OFBIZ_WRAPPER);
 
     public static BeansWrapper getDefaultOfbizWrapper() {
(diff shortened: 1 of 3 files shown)


framework/base/src/main/java/org/apache/ofbiz/base/util/template/FreeMarkerWorker.java
```diff
@@ -112,10 +115,10 @@ public static Configuration makeConfiguration(BeansWrapper wrapper) {
         Configuration newConfig = newConfiguration();
 
         newConfig.setObjectWrapper(wrapper);
-        TemplateHashModel rawStaticModels = wrapper.getStaticModels();
-        newConfig.setSharedVariable("Static", RestrictedStaticModels.fromConfig(rawStaticModels, "freemarker-whitelist"));
+        TemplateHashModel staticModels = getConfiguredStaticModel(wrapper);
+        newConfig.setSharedVariable("Static", staticModels);
         try {
-            newConfig.setSharedVariable("EntityQuery", rawStaticModels.get("org.apache.ofbiz.entity.util.EntityQuery"));
+            newConfig.setSharedVariable("EntityQuery", staticModels.get("org.apache.ofbiz.entity.util.EntityQuery"));
```
jacopoc: With this change, org.apache.ofbiz.entity.util.EntityQuery must be in the whitelist, which is not the case in the default configuration.

jacopoc (changes requested):
We could probably explicitly add EntityQuery to the allow-list to make your new code to work with default configuration.

framework/base/src/main/java/org/apache/ofbiz/base/util/template/FreeMarkerWorker.java
Criux: Good catch, thank you! I have reverted to the raw static models for EntityQuery so it is always included.

PR #984, opened 2026-03-14
Improved the code of ProgramExport and did the sandboxing/sanitising enhancements for the groovy script use.

ashishvijaywargiya:
Improved the code of ProgramExport and did the sandboxing/sanitising enhancements for the groovy script use. Used the SecureASTCustomizer to limit the user operations to the application and the database only.


golja:
CVE-2026-31388

PR #988, opened 2026-03-16
Fixed: Improve image upload handling by using file extensions for temporary files and ensuring safe copying of sanitized images

jacopoc:


golja:
CVE-2026-29226

PR #990, opened 2026-03-17
Fixed: Enhance temporary file handling for uploads by using original file extensions and ensuring safe copying of files

jacopoc:


golja:
CVE-2026-29226

PR #1033, opened 2026-03-26
Enable Develocity remote build cache

ribafish:
## Summary

Enable the Develocity remote build cache for all builds. Push is limited to CI (`GITHUB_ACTIONS`), so CI populates the cache and developers benefit from it on local builds.

The remote cache was explicitly disabled (`enabled = false`). This change sets `enabled = true` with `push = isCI`, following the standard Develocity configuration pattern.

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -44,7 +44,8 @@ buildCache {
     }
 
     remote(develocity.buildCache) {
-        enabled = false
+        enabled = true
+        push = isCI
     }
 }
 


ribafish:
Hi @JacquesLeRoux , thank you for the update. No rush, I completely understand it's not a pressing issue, just a quality of life improvement 👍

jacopoc:
@ribafish Now that I’ve merged your other pull request #1028, which made the `test` task cache-friendly, we could also merge this pull request. Currently, our CI [workflow](https://github.com/apache/ofbiz-framework/blob/trunk/.github/workflows/gradle.yaml) runs:

`./gradlew check javadoc`

However, our goal is to switch to the following as soon as we resolve some build issues:

`./gradlew pullAllPluginsSource check javadoc loadAll testIntegration`

My only concern is that merging this PR before we understand and fix the issues preventing us from running the new tasks might make it more difficult to debug why they fail on CI but succeed locally. Moreover, since `loadAll` is not currently cache-friendly, enabling it may again lead to a significant number of cache misses.

I’ll try to investigate these issues, but if you have any advice, please let me know. Thank you in advance!

PR #1043, opened 2026-03-31
Improved: introduce and use a new label for product receipts with zero accepted and rejected units (OFBIZ-13376)

golja:
This PR improves the message displayed when a receipt has both accepted and rejected quantities set to zero.


JacquesLeRoux:
Thanks Anihita, sounds quite good to me. I push both PRs and will test after (should not be any issue)

JacquesLeRoux:
Ah, I already pushed https://github.com/apache/ofbiz-framework/pull/1027

PR #1050, opened 2026-04-03
Fixed: Resolve test errors in framework-only deployments by adding spring-web dependency

jacopoc:
Added spring-web version 6.2.17 to dependencies.gradle.

SimpleMethodTest (in the main source set of framework/testtools) statically initializes MockHttpServletResponse, which at class-load time references org.springframework.http.MediaType from spring-web. spring-web was never declared as a direct dependency, it was only arriving transitively via the LDAP plugin's CAS dependency. Without any plugins, it's missing from the classpath.


JacquesLeRoux:
Hi Jacopo,

I found the same yesterday http://shibboleth.net/pipermail/commits/2015-October/022972.html
As it was a bit old I had a doubt.

PR #1055, opened 2026-04-03
Improved: Remove gradle-wrapper.jar from the repository and enhance the init-gradle-wrapper scripts to download and verify the version specified in gradle-wrapper.properties

jacopoc:
This change simplifies gradlew upgrades (no longer requiring updates to the scripts or recomputation of file hashes) and improves repository quality by removing the binary file (gradle-wrapper.jar). The behavior of init-gradle-wrapper.sh and init-gradle-wrapper.ps1 is now consistent.

The init-gradle-wrapper.sh and init-gradle-wrapper.ps1 scripts now:

* download only gradle-wrapper.jar
* parse the version from gradle-wrapper.properties (no hardcoded version in the scripts)
* fetch the SHA-256 from services.gradle.org and verify it
* correctly detect when the JAR already exists and fetch SHA, warn-and-continue if unreachable, exit if match, re-download if mismatch
* accept arguments:
** --help: prints usage + full upgrade workflow, then exits
** --upgrade: sets a flag that triggers "./gradlew wrapper" at both success exit points (jar-already-valid and freshly-downloaded-and-verified)
** unknown arguments produce an error with a pointer to --help


JacquesLeRoux:
I'll review and especially test, maybe today, else tmrw.
I'm currently blocked (Win7) and I'll maybe be blocked when testing, let's see...

JacquesLeRoux:
OK, as I supposed, I'm definitely blocked. Windows 7 supports only Gradle 8.11 and npm 13.14.0.
As it's my choice, I'll bypass the issue with a script (several actually), like I did for npm years ago.
It's great that you handled the init-gradle-wrapper.ps1 scripts!

BTW, I got another issue on Ubuntu 20.04 related to node in example. Cloning both repos fixed it.

JacquesLeRoux:
I let you handling https://github.com/apache/ofbiz-framework/pull/1052 ?

PR #1080, opened 2026-04-09
Upgrade the Gradle Wrapper used by release24.09 to version 8.14.4 (the same version as trunk) and fix code style issues

jacopoc:
By using the same version of the Gradle Wrapper in both trunk and release24.09, we can simplify the configuration of the CI/CD workflow.

Moreover, with the new version, the code style checks are more consistent between trunk and release24.09: this pull request  takes care of fixing the style issues on release24.09 by backporting the fixes from trunk.


JacquesLeRoux:
I totally agree

PR #1131, opened 2026-04-23
Fixed: sendInvoicePerEmail fails with MissingMethodException (OFBIZ-13388)

arashpreet0011:
JIRA: https://issues.apache.org/jira/browse/OFBIZ-13388

   Fixes the MissingMethodException thrown when sendInvoicePerEmail
   runs asynchronously. See JIRA for full stack trace and reproduction.

### DIFF
--- a/applications/accounting/src/main/groovy/org/apache/ofbiz/accounting/invoice/InvoiceServicesScript.groovy
+++ b/applications/accounting/src/main/groovy/org/apache/ofbiz/accounting/invoice/InvoiceServicesScript.groovy
@@ -326,13 +326,13 @@ Map cancelInvoice() {
  * @return Success response
  */
 Map sendInvoicePerEmail() {
+    Map ctx = [*: parameters,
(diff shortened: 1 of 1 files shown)


applications/accounting/src/main/groovy/org/apache/ofbiz/accounting/invoice/InvoiceServicesScript.groovy
```diff
@@ -564,7 +563,7 @@ Map addTaxOnInvoice() {
         addTaxMap.itemAmountList << totalAmount
         addTaxMap.itemPriceList << it.amount
         addTaxMap.itemQuantityList << it.quantity
-        addTaxMap.itemShippingList << BigDecimal.ZERO
+        addTaxMap.itemShippingList << 0
```
vyasdevanshu, line 566: This change is out of context of the ticket fix. 
Is there any specific reason to include this in the current changes?

PR #1159, opened 2026-05-04
Fixed: Add file path verifications to Webtools Entity Import/Export and improve redirection checks for URL DataResources

jacopoc:
Many thanks to icysun (icysun@qq.com) for the thorough analysis and valuable suggestions to improve the robustness of Entity data import/export processing and URL DataResource rendering.


icysun:
thanks

PR #1168, opened 2026-05-06
OFBIZ-13262: Added inventory reservation logic for production run tasks in manufacturing and also improved inventory issuance feature that works with reservations

diveshdut:
(OFBIZ-13262)

This PR introduces formal Inventory Reservation capabilities to the Manufacturing module. Previously, production components were largely tracked via direct issuance; this enhancement allows the system to reserve components ahead of physical issuance, bringing manufacturing parity with Sales Order fulfillment workflows.

**User Workflow & UI Journey**

1. Creation & Planning: A Production Run is created with estimated component requirements.

2. Formal Commitment (Approval): Upon changing the status to Printed or Approved, the system automatically triggers the reservation engine (if setting is enabled)

3. UI Impact: The "Components" screen now displays a "Reserved" column and a "Backordered" field for shortages.

4. Seamless Issuance: When the user clicks "Issue Component," the system automatically consumes the existing reservations first, ensuring that reserved lots or batches are accurately tracked through the issuance process. If no reservations are available, then system issues inventory. 

5. "Force Issue" (Break-Glass Logic): In cases where reservations are not available (e.g., emergency jobs or stock discrepancies), the user can still perform a "Force Issue." This allows physical issuance to proceed regardless of the reservation state, ensuring that the production line never stops due to administrative blockers.

6. Automated Recovery: Any remaining backorders in the reservation ledger are automatically satisfied as soon as new stock is received in the warehouse.

**New Data Model: WorkEffortInvRes**

A new table, **WorkEffortInvRes** (Inventory Reservation for WorkEffort), has been introduced as the core ledger for this feature.

- Purpose: It tracks "soft" commitments of materials to production tasks.

- Backorder Support: Through the quantityNotAvailable field, it explicitly tracks "Global Debt" or backordered units, allowing the system to know exactly what is owed to the production line.

- Auditability: It stores reservation timestamps and strategy enums (FIFO/LIFO), providing a full audit trail of how and when materials were committed.

**New Policy Controls (Facility Settings)**

To support diverse warehouse operations, three new flags have been introduced to the Facility entity:

1. autoReservePrun (Y/N): Controls the entry point of the reservation lifecycle. If enabled, the system automatically invokes autoReserveWorkEffortInventory whenever a production run reaches the Approved or Printed status.

7. allowInventoryTheft (Y/N): Determines if a production task can "claim" reservations currently held by another task. Essential for high-priority production in fluid environments.

8. reconcilePrunBackorders (Y/N): Enables real-time backorder fulfillment. When inventory is received, the system proactively reserves it for pending production tasks.

**Core Services Added:**

1. autoReserveWorkEffortInventory: Triggers when autoReservePrun in Facility is set to Y.  Provides high-level automation by orchestrating the reservation of all components required for a production run. It evaluates current facility inventory levels and automatically generates WorkEffortInvRes records to secure materials.

2. reserveWorkEffortInventoryItem: Allows for granular, item-specific commitments. Users can explicitly reserve a particular InventoryItem (e.g., a specific batch or lot) against a production task, bypassing the automated selection logic.

3. reallocateAndIssueInventory: The primary engine for transition. It manages the logic of converting existing reservations into physical issuance. It also handles "Move-Then-Issue" strategies, allowing reservations to be reallocated between tasks if required by the facility's fluidity policies. Depends on allowInventoryTheft flag of Facility.

9. releaseProductionRunTaskComponent: Handles the cleanup of commitments. When a task is cancelled or a component is no longer needed, this service releases the reservation and ensures that the ATP (Available to Promise) balance is immediately restored across the facility.

10. reassignWorkEffortInventoryReservations: The Backorder Satisfier: Automatically fulfills backordered production tasks (quantityNotAvailable in WorkEffortInvRes) when new inventory is received or released.

11. reconcileGlobalReservations: The Public Auditor: The top-level service used to audit and reconcile inventory state. It ensures that any "Global Debt" (reservations without an item ID) is correctly localized to physical stock as soon as it becomes available.

12. trinitySync: The Integrity Guard: Performs real-time recalculation of availableToPromiseTotal by comparing physical stock against active reservations in WorkEffortInvRes and OrderItemShipGrpInvRes.

**Real-time Automation (SECA)**

To ensure a seamless user experience, several Service ECA (SECA) triggers have been introduced:

- Approval Trigger: Automatically invokes autoReserveWorkEffortInventory when a production run status changes to Printed/Approved, ensuring stock is secured the moment a job is confirmed.

- Receipt Satisfaction: Automatically invokes reassignWorkEffortInventoryReservations whenever inventory levels increase (via balanceInventoryItems), immediately satisfying any pending production backorders.

The primary objective is to allow production tasks to commit inventory early in the production lifecycle. This ensures that critical materials are "locked" for specific jobs, preventing stockouts and enabling more accurate Material Requirements Planning (MRP) and Available-to-Promise (ATP) calculations.

**Business Impact**

- Accurate Material Planning: MRP and ATP calculations now reflect true manufacturing commitments, preventing component "hoarding" and stockouts.

- Zero-Touch Fulfillment: Production backorders are satisfied automatically upon stock receipt, reducing manual intervention from warehouse managers.

- Policy Flexibility: Facility managers can choose between strict reservation models or fluid environments where high-priority jobs can "steal" stock from others.

At a high level, this new inventory reservation framework delivers five critical business benefits:

1. Guaranteed Material Availability: By formally "locking" components for specific production tasks, the system prevents stockouts and ensures that materials are physically waiting for the production line when needed.

2. Reduced Production Lead Times: The automated backorder fulfillment logic removes manual intervention from the warehouse. As soon as missing stock arrives, it is instantly committed to the waiting production job, accelerating the entire manufacturing cycle.

3. Physical-to-Ledger Integrity: The "Trinity of Truth" synchronization ensures that your warehouse floor always matches your financial books. This eliminates "ghost inventory" and significantly reduces the time spent on manual inventory audits and corrections.

4. Strategic Operational Agility: New facility-level policies allow management to prioritize critical production runs. High-priority jobs can automatically secure materials from lower-priority tasks, ensuring that key customer deadlines are always met.

5. Proactive Shortage Visibility: Floor managers gain real-time visibility into material blockers. By identifying backorders at the planning stage rather than the issuance stage, the organization can proactively resolve shortages before they stop the production line.

In essence, these features move the organization from reactive inventory tracking to proactive material commitment, directly increasing manufacturing reliability and throughput.

### DIFF
--- a/applications/accounting/minilang/ledger/GeneralLedgerServices.xml
+++ b/applications/accounting/minilang/ledger/GeneralLedgerServices.xml
@@ -1635,6 +1635,19 @@ under the License.
             <field-map field-name="inventoryItemId" from-field="parameters.inventoryItemId"/>
         </entity-one>
         <get-related-one value-field="workEffortInventoryAssign" relation-name="InventoryItem" to-value-field="inventoryItem"/>
+        <set field="partyAccountingPreferencesCallMap.organizationPartyId" from-field="inventoryItem.ownerPartyId"/>
+        <call-service service-name="getPartyAccountingPreferences" in-map-name="partyAccountingPreferencesCallMap">
+             <result-to-field result-name="partyAccountingPreference" field="partyAcctgPreference"/>
+        </call-service>
+        <if>
+            <condition>
+                <if-compare field="partyAcctgPreference.enableAccounting" value="N" operator="equals"/>
+            </condition>
+            <then>
+                <log level="warning" message="The internal organization with id [${partyAccountingPreferencesCallMap.organizationPartyId}] has no PartyAcctgPreference setting; the following accounting transaction will be ignored."/>
+                <return/>
+            </then>
+        </if>
         <!-- TODO: handle serialized inventory -->
         <calculate field="origAmount" decimal-scale="${ledgerDecimals}" rounding-mode="${roundingMode}">
             <calcop operator="multiply">
@@ -1660,6 +1673,19 @@ under the License.
         <set field="creditEntry.origAmount" from-field="origAmount"/>
         <set field="creditEntry.origCurrencyUomId" from-field="inventoryItem.currencyUomId"/>
         <set field="acctgTransEntries[]" from-field="creditEntry" type="Object"/>
+
+        <!-- An inventory item detail record is created to keep track of the units (value) that we are going to add to the inventory account -->
+        <set field="quantityToDeduct" from-field="parameters.quantity" type="BigDecimal"/>
+        <if-empty field="quantityToDeduct">
+            <log level="warning" message="Accounting issuance for WorkEffort [${parameters.workEffortId}] and InventoryItem [${parameters.inventoryItemId}] is missing the incremental 'quantity' parameter. Falling back to assignment total: ${workEffortInventoryAssign.quantity}"/>
+            <set field="quantityToDeduct" from-field="workEffortInventoryAssign.quantity" type="BigDecimal"/>
+        </if-empty>
+
+        <set field="createInventoryDetailMap.inventoryItemId" from-field="parameters.inventoryItemId"/>
+        <set field="createInventoryDetailMap.workEffortId" from-field="parameters.workEffortId"/>
+        <set field="createInventoryDetailMap.accountingQuantityDiff" value="${(-1) * quantityToDeduct}" type="BigDecimal"/>
+        <call-service service-name="createInventoryItemDetail" in-map-name="createInventoryDetailMap"/>
+
         <!-- Set header fields (AcctgTrans) -->
         <set field="createAcctgTransAndEntriesInMap.glFiscalTypeId" value="ACTUAL"/>
         <set field="createAcctgTransAndEntriesInMap.acctgTransTypeId" value="INVENTORY"/>
--- a/applications/accounting/servicedef/services_ledger.xml
+++ b/applications/accounting/servicedef/services_ledger.xml
@@ -411,6 +411,7 @@ under the License.
         <description>Create an accounting transaction for inventory that is issued to a work effort (Type: INVENTORY D: RAWMAT_INVENTORY, C: WIP_INVENTORY)</description>
         <attribute name="workEffortId" type="String" mode="IN" optional="false"/>
         <attribute name="inventoryItemId" type="String" mode="IN" optional="false"/>
+        <attribute name="quantity" type="BigDecimal" mode="IN" optional="true"/>
         <attribute name="acctgTransId" type="String" mode="OUT" optional="true"/>
     </service>
     <!-- Service for the automatic creation of AcctgTransForPhysicalInventoryVariance, triggered by SECAs -->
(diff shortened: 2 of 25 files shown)


jacopoc:
@diveshdut thank you for your contribution.

I have a couple of concerns.

The **first one** is minor and regards the use of words like "theft" and "stealing" to metaphorically represent concepts that could instead be named in a more formal, standard, or traditional way (e.g., "inventory reallocation" or something similar). In my opinion, this would make the system more generic and neutral and would prevent any risk of confusion. The same applies to "trinitySync".

The **second one** concerns the introduction of the "trinitySync" service. From your description in the dev list email (quoted below), it seems that you are describing a bug that needs to be addressed: the system processes should never bring the data into the inconsistent state you described. Introducing a service to fix the data afterward does not seem like the right approach to me. Once the core issue causing the data inconsistency is resolved, this new "feature" ("Trinity Sync") would not even be required.

> *5. Architectural Stability: The "Trinity Sync"*
> One of the core problems we identified was "Ledger Blindness." Since the
> InventoryItem table does not have a quantityNotAvailable (QNA) field, the
> availableToPromiseTotal (ATP) can easily drift if reservations are modified
> without corresponding ledger entries.
> 
> For example, let's take the Force Issuance (Theft) scenario:
> 
> 
> *Initial State:*
> Inventory: QOH=5, ATP=0.
> 
> Task A: Has a physical reservation of 5 units (secured).
> 
> Task B: Has a backordered reservation of 5 units (QNA=5).
> 
> *The Action: *A floor manager performs a "Force Issue" for Task B, stealing
> the 5 units from Task A.
> 
> 
> *The Transformation:*
> Task B now has 5 physical units issued/committed.
> 
> Task A's reservation is automatically converted from "Physical" to
> "Backordered" (QNA becomes 5).
> 
> 
> *The Drift:*
> The Ledger Error: During this "theft," the InventoryItem ledger can easily
> drift. If the system "releases" Task A's reservation to give it to Task B,
> the ATP might momentarily jump to 5. If the satisfaction of Task B's
> backorder doesn't precisely trigger a corresponding ledger subtraction at
> the exact same moment, the ATP remains at 5.
> 
> *The Result: *The system incorrectly reports ATP=5, even though all 5 units
> are physically issued to Task B and 5 units are still owed to Task A. You
> have created 5 units of "Ghost Inventory."
> 
> *The Solution:* The trinitySync service solves this by performing a cold
> recalculation of the "Physical Truth." It looks at the resulting state of
> all reservations after the theft, determines that True ATP must be 0, and
> forces the ledger to match. This ensures that a "Force Issuance" for one
> task doesn't accidentally inflate the availability for others.
> 
> So this auditor service performs a three-way check between:
> 
> Physical Stock (QOH)
> 
> Logical Commitments (Active Reservations in WorkEffortInvRes and
> OrderItemShipGrpInvRes)
> 
> The Promise Ledger (ATP)
> 
> If a drift is detected, trinitySync calculates the "True ATP" and applies a
> correction via the createInventoryItemDetail service, ensuring the ledger
> always reflects physical reality and maintains financial auditability.

diveshdut:
@jacopoc thanks for your review and comment. I agree with you. I am working to fix the deeper bug and remove the use of "trinitySync" service. I will also work to rename service and comments from stealing/theft to reallocation or some thing similar.

diveshdut:
@jacopoc 

Thank you for the review and for pointing this out. Your concern was valid: fixing the data afterward through the trinitySync service was not the right approach.

I have removed the trinitySync service usage and updated the core inventory issuance/reallocation logic so the inconsistent state is handled at the source instead of being corrected afterward.

I also renamed the “theft/stealing” terminology to neutral “inventory reallocation” wording across the code, tests, labels, and related facility policy field names.

PR #1199, opened 2026-05-12
Fixed: Enable indirect import checks and remove Object from receivers whitelist in ProgramExport for enhanced security

jacopoc:


golja:
CVE-2026-45434

PR #1212, opened 2026-05-15
Fixed: error on the find party screen. (OFBIZ-10824)

rehan3586:
Fixed:error on the find part screen.
(OFBIZ-10824)


rehan3586:
When searching for a party using both Postal Address and Telecom Number criteria simultaneously, an SQL Exception is thrown: "Column name 'PC.PARTY_ID' is in more than one table in the FROM list."
 
This was caused by the `PartyContactMech` member entity being added to the `DynamicViewEntity` multiple times with the exact same alias ("PC") when dynamically building the query in the `performFindParty` service.
 
The fix is to use unique aliases for the `PartyContactMech` entity depending on the contact mech type being joined (e.g., "PCPA" for PostalAddress, "PCCM" for generic ContactMech, "PCTM" for TelecomNumber). This prevents alias collisions in the generated SQL statement when multiple contact mechanisms are searched at once.

PR #1230, opened 2026-05-21
Fixed: Correct the Unix command for downloading the Gradle wrapper in README (OFBIZ-13417)

visvendra1991:
Fixed: Correct the Unix command for downloading the Gradle wrapper in README

(OFBIZ-13417)

### Explanation

The "Quick start" section of `README.md` documented the Unix command to download the Gradle wrapper as an absolute path:

```
Unix-like OS: `/gradle/init-gradle-wrapper.sh`
```

The leading slash makes it an absolute path, so running it from the OFBiz home directory fails with `No such file or directory`.

This changes it to a relative path with the `sh` prefix:

```
Unix-like OS: `sh gradle/init-gradle-wrapper.sh`
```

which matches the script's own `--help` usage text (`gradle/init-gradle-wrapper.sh` prints `Usage: sh gradle/init-gradle-wrapper.sh [--help] [--upgrade]`) and is consistent with the Windows command shown just above it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,7 +106,7 @@ MS Windows: `init-gradle-wrapper`
 > **Note** -
 > If you wonder where are stored the PowerShell Executables, here are the answers: [https://s.apache.org/w5dye](https://s.apache.org/w5dye)
 
-Unix-like OS: `/gradle/init-gradle-wrapper.sh`
+Unix-like OS: `./gradle/init-gradle-wrapper.sh`
 
 ### Prepare OFBiz
 


README.md
```diff
@@ -106,7 +106,7 @@ MS Windows: `init-gradle-wrapper`
 > **Note** -
 > If you wonder where are stored the PowerShell Executables, here are the answers: [https://s.apache.org/w5dye](https://s.apache.org/w5dye)
 
-Unix-like OS: `/gradle/init-gradle-wrapper.sh`
+Unix-like OS: `sh gradle/init-gradle-wrapper.sh`
```
jacopoc: For consistency with other *nix commands described in the README, it would be better to use instead:
`./gradle/init-gradle-wrapper.sh`

jacopoc (changes requested):

README.md
visvendra1991: Thanks @jacopoc, agreed — that's more consistent with the other Unix-like commands in the README. Updated to `./gradle/init-gradle-wrapper.sh`.

jacopoc:
Thank you @visvendra1991, I have merged your Pull Request!

PR #1254, opened 2026-05-22
Fixed reservation logic for Auto Inventory allocation process. OFBIZ-13386

nameet-jain:
Fixed reservation logic for Auto Inventory allocation process. OFBIZ-13386

### DIFF
--- a/applications/order/src/main/java/org/apache/ofbiz/order/shoppingcart/ShoppingCart.java
+++ b/applications/order/src/main/java/org/apache/ofbiz/order/shoppingcart/ShoppingCart.java
@@ -671,7 +671,7 @@ public class ShoppingCart implements Iterable<ShoppingCartItem>, Serializable {
             }
             item = ShoppingCartItem.makeItem(0, productId, selectedAmount, quantity, null,
                     reservStart, reservLength, reservPersons, accommodationMapId, accommodationSpotId, shipBeforeDate, shipAfterDate,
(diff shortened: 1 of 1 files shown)


jacopoc:
Thank you, @nameet-jain.

Could you please provide the steps you followed to test the fix?

I’m having a hard time understanding the expected workflow from the descriptions provided in [OFBIZ-10518](https://issues.apache.org/jira/browse/OFBIZ-10518) and [OFBIZ-13386](https://issues.apache.org/jira/browse/OFBIZ-13386)

nameet-jain:
**Steps to Reproduce**

1. Go to Catalog → Stores and select any store (for example, 9000).
2. Set the Allocate Inventory flag to Y. This enables the Auto Reserve flag on the order entry screen.
3. Place an order from Order Manager with the Auto Reserve flag checked.

Actual Result

- The order is placed successfully, but the inventory is not reserved. You can verify this on the Order View screen.
- Also navigate to Order → Allocation Plan → Find All Allocation Plans. The Allocation Plan method is shown as Manual, which indicates that the reservation is planned and must be processed manually through the Allocation Plan screens.

Expected Result

- The order should be placed successfully, and the inventory should be reserved automatically.
- The system creates the Allocation Plan with the Auto method so that the planned allocation is processed automatically.

@jacopoc , here is another [PR](https://github.com/apache/ofbiz-framework/pull/1237) that fixes the display value of the Allocate Inventory flag. The default behavior for this flag should be N, but it is currently displayed as Y when the database value is empty (null), which is misleading.

PR #1285, opened 2026-05-27
Improved: Implement Type safe Config in OFBiz

nmalin:


framework/base/src/main/java/org/apache/ofbiz/base/config/AbstractConfigElement.java
```diff
@@ -0,0 +1,100 @@
...
+import org.apache.ofbiz.entity.GenericEntityConfException;
+import org.w3c.dom.Element;
+
+/**
+ * Interface that must be implemented by every class wishing to use the centralized configuration reader tool
+ * Each class implementing this interface <b>must</b> :
+ * <ul>
+ *     <li>Possess a <code>public static final String ELEMENT_NAME</code> containing the name of the XML element to which the class refers to</li>
+ *     <li>If needed a <code>public static final String ELEMENT_FIELD_ID_NAME</code> containing the attribut name of the XML element
+ *         that identify as unique the element. Else the <code>name</code> string will be used</></></li>
```
jacopoc: The above line is causing a javadoc error (malformed HTML) due to the presence of `</></>`

framework/base/src/main/java/org/apache/ofbiz/base/config/AbstractConfigElement.java
```diff
@@ -0,0 +1,100 @@
...
+ *     <li>Have a first constructor with an <code>Element</code> and a <code>String</code> as parameters</li>
+ *     <li>Have a second constructor with a <code>Map&lt<span/>String, Object&gt</code> and a <code>String</code> as parameters</li>
```
jacopoc: The above line is causing some javadoc errors (self-closing element not allowed, semicolon missing, etc...).

framework/base/src/main/java/org/apache/ofbiz/base/config/AbstractConfigElement.java
nmalin: Yes right jacopo, I only run ./gradlew check on local and not javdoc :sweat_smile: I will fix that

framework/base/src/main/java/org/apache/ofbiz/base/config/AbstractConfigElement.java
nmalin: I fixed it, thanks for this spot !

PR #1294, opened 2026-05-28
Fixed: Sanitize fetched page text in HttpViewHandler

mridulpathak:
Fixed: Sanitize fetched page text in HttpViewHandler

### DIFF
--- a/framework/webapp/src/main/java/org/apache/ofbiz/webapp/view/HttpViewHandler.java
+++ b/framework/webapp/src/main/java/org/apache/ofbiz/webapp/view/HttpViewHandler.java
@@ -28,6 +28,7 @@ import jakarta.servlet.http.HttpServletResponse;
 import org.apache.ofbiz.base.util.Debug;
 import org.apache.ofbiz.base.util.HttpClient;
 import org.apache.ofbiz.base.util.HttpClientException;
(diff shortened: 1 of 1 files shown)


jacopoc:
@mridulpathak My guess, after reviewing the changes in this pull request, is that you are trying to address some "code quality" alerts reported by a static analysis tool (such as CodeQL). Is my guess correct?

If that is the case, my suggestion would be to postpone the decision to merge it, even if it resolves the alerts, provided that the changes do not actually change the behavior of the system (i.e., they are functionally equivalent to the current version) and only make the code easier for static analysis tools to assess.

I suggest this because I have found a few other alerts that *may* be false positives and *may* also be resolved by simplifying the code path. For now, I have decided not to merge changes that are functionally equivalent to the current code solely to resolve alerts, and instead focus only on resolving alerts that require functional changes to the codebase.

Once all valid alerts are resolved, we can reconsider the ones that appear to be false positives and decide how we want to address them (possibly revisiting pull requests like this).

What do you think?

mridulpathak:
Sure @jacopoc. It makes complete sense. I'll hold on to this one and focus on identifying and fixing valid alerts. We could revisit these false positives later.

PR #1311, opened 2026-06-02
Removed: Deprecated EntityDateFilterCondition references from EntityConditionVisitor and EntityConditionVisitorTests (OFBIZ-3880)

mridulpathak:
Removed: Deprecated EntityDateFilterCondition references from EntityConditionVisitor and EntityConditionVisitorTests (OFBIZ-3880)


jacopoc:
@mridulpathak, without having looked at the details of your work, I was expecting you to deprecate the following method in the `EntityConditionVisitor` interface:
`void visit(EntityDateFilterCondition df);`
and replace it with a new one. Instead, you removed it. 
Maybe this is totally fine, but please have a look.

mridulpathak:
@jacopoc In the new implementation `EntityDateFilterCondition` is completely deprecated and the respective methods have been moved to `EntityCondition` class which is already covered by `EntityConditionVisitor` interface.

PR #1314, opened 2026-06-02
Improved Move REST-API plugin into framework (OFBIZ-13415)

Lukas-Finster:
Migrates the REST-API plugin into the core framework. Additionally 'docs' folder was renamed to 'swagger' to be less generic


dixitdeepak:
Thanks @Lukas-Finster.

We should check the possibility of migrating the plugin while preserving its history. Git provides mechanisms to do this, but we need to verify the ASF-approved approach and ensure it aligns with ASF project practices.

Lukas-Finster:
Good idea @dixitdeepak,
I will close the PR for now and check on that!

PR #1322, opened 2026-06-02
Fixed: Correct receivedQuantity calculation in getReceivedQuantityForOrderItem by including quantityRejected (OFBIZ-13429)

golja:
As described in [OFBIZ-13429](https://issues.apache.org/jira/browse/OFBIZ-13429), when the 'Quantity Received' value exceeds the 'Quantity' specified on a purchase order, the system automatically updates the related issuance, shipment, and order items through the updateIssuanceShipmentAndPoOnReceiveInventory shipment receipt service.

However, when one or more receipts contain rejected units, order and shipment updates fail. The issue is caused by a bug in the getReceivedQuantityForOrderItem() function, which is invoked by the updateIssuanceShipmentAndPoOnReceiveInventory service to determine the received quantity.

This PR fixes the issue by including rejected units in the quantity calculation performed by getReceivedQuantityForOrderItem(), which previously considered only accepted units and ignored rejected quantities.

### DIFF
--- a/applications/product/src/main/groovy/org/apache/ofbiz/product/shipment/ShipmentReceiptServices.groovy
+++ b/applications/product/src/main/groovy/org/apache/ofbiz/product/shipment/ShipmentReceiptServices.groovy
@@ -299,6 +299,7 @@ BigDecimal getReceivedQuantityForOrderItem (GenericValue orderItem) {
     List shipmentReceipts = from('ShipmentReceipt').where(orderId: orderItem.orderId, orderItemSeqId: orderItem.orderItemSeqId).queryList()
     for (GenericValue shipmentReceipt : shipmentReceipts) {
         receivedQuantity +=  shipmentReceipt.quantityAccepted
+        receivedQuantity +=  shipmentReceipt.quantityRejected
     }
     return receivedQuantity
 }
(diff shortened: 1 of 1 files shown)


nmalin:
Hi @golja 

Thanks for this !

Little question, quantityRejected is never empty ?

This function would be simplifying like this
```java
BigDecimal getReceivedQuantityForOrderItem (GenericValue orderItem) {
    return from('ShipmentReceipt')
            .where(orderId: orderItem.orderId,
                    orderItemSeqId: orderItem.orderItemSeqId)
            .queryList()
            .sum { it.quantityAccepted + it.quantityRejected }
}
```
or is quantityRejected can be empty
```java
BigDecimal getReceivedQuantityForOrderItem (GenericValue orderItem) {
    return from('ShipmentReceipt')
            .where(orderId: orderItem.orderId,
                    orderItemSeqId: orderItem.orderItemSeqId)
            .queryList()
            .sum { it.quantityAccepted + (it.quantityRejected ?: 0) }
}
```

golja:
Hi @nmalin, 
this is definitely better than the current code, thanks! I'll update my PR accordingly.

golja:
Hi @nmalin,
I've checked that both quantityAccepted and quantityRejected can potentially be null, using the query
```
Select COLUMN_NAME, IS_NULLABLE
from INFORMATION_SCHEMA.COLUMNS
where TABLE_NAME = 'SHIPMENT_RECEIPT'
```
Thus, I've also added a check for quantityAccepted.

Thanks again,

Anahita

nmalin:
:) you're welcome

PR #1332, opened 2026-06-04
Handle trailing empty CSV headers in invoice import and fix Accounting ImportExport menu navigation.(OFBIZ-12089)

ratneshup:
Fixed:  Invoice import is not visible from UI and not working (OFBIZ- 12089) (#1332)

Prevented IllegalArgumentException and NullPointerException during CSV imports when uploaded templates contain trailing empty column headers (trailing commas). Also fixed Accounting Import/Export menu visibility and navigation.


JacquesLeRoux:
Thanks Ratnesh!

PR #1340, opened 2026-06-05
Fixed: Enhance DataResource permission checks to include dataTemplateTypeId validation when updating an existing record

jacopoc:


golja:
CVE-2026-50223

PR #1345, opened 2026-06-10
Implemented: Add new decorator FilteredListDecorator

nmalin:
A new type of decorator has been added to handle searches at the top of lists.

Currently, we have the FindScreenDecorator, which allows us to define a page structure for searching, comprising a search form and a results form. This new decorator offers the same functionality but is designed to be more compact. It therefore enables a compact search form linked to a list, acting as a filter. The decorator’s structure allows for direct management of dynamic refreshing.

This new structure is particularly useful in sub-screen lists, enabling lightweight searches.

An example has been added to the WorkEffort compounds, where, in the party tab linked to a WorkEffort, you can quickly filter by status or role.

Capture d’écran du 2026-06-13 16-39-40
Capture d’écran du 2026-06-13 16-39-58

### DIFF
--- a/applications/workeffort/widget/CommonScreens.xml
+++ b/applications/workeffort/widget/CommonScreens.xml
@@ -213,6 +213,24 @@ under the License.
         </section>
     </screen>
 
+    <screen name="FilteredListDecorator">
+        <section>
+            <actions>
+                <property-map resource="WorkEffortUiLabels" map-name="uiLabelMap" global="true"/>
+                <property-map resource="OrderUiLabels" map-name="uiLabelMap" global="true"/>
+                <property-map resource="PartyUiLabels" map-name="uiLabelMap" global="true"/>
+                <property-map resource="CommonUiLabels" map-name="uiLabelMap" global="true"/>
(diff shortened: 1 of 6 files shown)


jacopoc:
Hi @nmalin,

Thank you for sharing this contribution.

The descriptions in the pull request and the associated commit are not entirely clear to me. Would it be possible to rephrase them more clearly?

If you prefer, feel free to write it in French and then use a modern machine translation tool to translate it into English.

Thanks in advance!

nmalin:
Hi @jacopoc, a kind message to confirm that my english skill isn't enough, and you right ! :sweat_smile: 

I took more time to rewrite the pr comment, I hope it's more understandable for you.  

Thanks for your time

nmalin:
And I forgot, any suggest on a better name for this decorator will be welcome, we try to do the best with our prism, but from another view, another idea :)

jacopoc:
Thank you, @nmalin. It all makes sense now!

I like the name you've chosen (`FilteredListDecorator`) because it describes its purpose well.

nmalin:
Thanks for you look Jacopo !

PR #1351, opened 2026-06-15
OFBIZ-13402 Move Groovy test sources from src/main to src/test

toaditi:
## Summary

Moves 52 Groovy test files from `src/main/groovy/` to `src/test/groovy/` across all active components, preventing test code from being compiled into production JAR files and distributed in release artifacts.

**Components affected:**
- `applications/accounting` (18 files)
- `applications/content`, `applications/marketing` (1 file each)
- `applications/order` (7 files)
- `applications/party` (7 files)
- `applications/product` + shipment (13 files)
- `framework/base`, `framework/service`, `framework/webapp` (3 files)
- `plugins/assetmaint`, `plugins/ecommerce`, `plugins/example` (3 files)

**`build.gradle` changes:**
- `createOfbizCommandTask` now appends `sourceSets.test.runtimeClasspath` when the task name contains `--test` or `-t`, so the OFBiz integration test runner can find the relocated classes at runtime.
- Added `src/test/groovy` to the `createPlugin` directory scaffold so new plugins start with the correct layout.

**Not moved:** `framework/service/src/main/groovy/.../TestServices.groovy` — its own comment states "this class is not for test, it's only services."

**XML test definition files** do not require changes — they reference classes by Java package name, which is unchanged.

## Test plan

- [x] Run `./gradlew compileGroovy compileTestGroovy` — should compile cleanly
- [x] Run `./gradlew checkstyleMain codenarcMain codenarcTest` — should pass
- [x] Run OFBiz integration tests (`./gradlew ofbiz --test`) — test classes should be discoverable via the updated classpath
- [x] Confirm production JAR no longer contains test class bytecode


gtchaboussie:
Hi @toaditi, it looks like a good idea to me, it was quite confusing to have test code in `src`.
No problems for me after code review, looks good.

PR #1354, opened 2026-06-16
Reverted the service definition verb attribute introduced by OFBIZ-11328 for mapping exported services to specific HTTP methods.

dixitdeepak:
- Exported services are currently exposed through the service engine and have historically been available via POST requests.
- Introducing HTTP method definitions in service descriptors adds REST-specific concerns to service definitions.
- The framework's REST implementation already provides a dedicated mechanism for defining REST endpoints through rest.xml.
- Service definitions should remain transport-agnostic and not require HTTP method metadata.
- Mixing service contracts and REST endpoint definitions creates duplication and can lead to inconsistent API behavior.

- Remove the action attribute from service definitions.
- Remove HTTP method mapping logic introduced for exported services.

Applications requiring REST-style endpoint definitions should use rest.xml and the REST framework rather than annotating service definitions with HTTP-specific metadata. This keeps service definitions independent of protocol concerns and preserves backward compatibility with existing exported service behavior.


jacopoc:
I agree with the removal of this old feature.

PR #1362, opened 2026-06-17
Fixed: Added missing component-load entry for rest-api (OFBIZ-13415)

Lukas-Finster:


Lukas-Finster:
During the revert of rest-api migration, the entry in component-load.xml was reverted as well, but never reapplied.

PR #1396, opened 2026-06-30
Fixed: Incorrect unit price calculation during PO receiving under non-US locales (OFBIZ-13449)

toaditi:
Jira: https://issues.apache.org/jira/browse/OFBIZ-13449

## Problem
Receiving a purchase order under a locale whose decimal separator is a comma (e.g. `pl_PL`) corrupts the order item unit price: `7,41` is stored as `741`.

## Root cause
The receiving form submits the order-currency unit price as a locale-formatted string. OFBiz's generic String→BigDecimal conversion (`ObjectType.simpleTypeOrObjectConvert` → `NumberConverters.fromString`) **ignores the caller locale** and falls back to `Locale.getDefault()` unless the `testBigDecimal` system property is set:

```java
// framework/base/.../conversion/NumberConverters.java
locale = null != System.getProperty("testBigDecimal") ? locale : Locale.getDefault();
```

So `simpleTypeOrObjectConvert("7,41","BigDecimal",...,pl_PL,...)` returns `741` (comma read as grouping), even though raw `NumberFormat.getNumberInstance(pl_PL).parse("7,41")` returns `7.41`. **Declaring the service attribute as `BigDecimal` does not help**, because the same converter runs — verified by integration test (it stored `741.000`).

## Fix
Parse the submitted order-currency unit price explicitly with the request locale (`parameters.locale`) in `updateIssuanceShipmentAndPoOnReceiveInventory`, bypassing the generic converter.

## Verification
Added `ReceivePoLocaleTest` (registered in `FacilityTest.xml`) that drives the service in a real container under `pl_PL` with `orderCurrencyUnitPrice="7,41"` and asserts the stored `unitPrice == 7.41`:
- Before this change: `741.000` (fails).
- After: `[JUNIT] Pass: true | # Tests: 1 | # Failed: 0`.

`checkstyleMain` / `codenarcMain` / `codenarcTest` pass.

## Note
The underlying `NumberConverters.fromString` behaviour (discarding the caller locale) looks like a broader framework issue and is probably worth a separate ticket; this PR fixes the receiving flow without changing that shared converter.

### DIFF
--- a/applications/product/src/main/groovy/org/apache/ofbiz/product/shipment/ShipmentReceiptServices.groovy
+++ b/applications/product/src/main/groovy/org/apache/ofbiz/product/shipment/ShipmentReceiptServices.groovy
@@ -20,6 +20,8 @@ package org.apache.ofbiz.product.shipment
 
 import java.math.RoundingMode
 import java.sql.Timestamp
+import java.text.DecimalFormat
+import java.text.NumberFormat
 
 import org.apache.ofbiz.base.util.UtilDateTime
 import org.apache.ofbiz.base.util.UtilProperties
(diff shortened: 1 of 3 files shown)


applications/product/src/main/groovy/org/apache/ofbiz/product/shipment/ShipmentReceiptServices.groovy
```diff
@@ -308,8 +325,9 @@ BigDecimal getReceivedQuantityForOrderItem (GenericValue orderItem) {
 Map updateIssuanceShipmentAndPoOnReceiveInventory() {
     GenericValue orderItem = from('OrderItem').where(parameters).queryOne()
     if (parameters.orderCurrencyUnitPrice) {
-        if (parameters.orderCurrencyUnitPrice != orderItem.unitPrice) {
-            orderItem.unitPrice = new BigDecimal (parameters.orderCurrencyUnitPrice)
+        BigDecimal orderCurrencyUnitPrice = parseLocalizedAmount((String) parameters.orderCurrencyUnitPrice, (Locale) parameters.locale)
+        if (orderCurrencyUnitPrice != orderItem.unitPrice) {
+            orderItem.unitPrice = orderCurrencyUnitPrice
             orderItem.store()
```
dixitdeepak, line 313: @toaditi 
ObjectType.simpleTypeOrObjectConvert also handles this conversion.
`price = (BigDecimal) ObjectType.simpleTypeOrObjectConvert(parameters.orderCurrencyUnitPrice, "BigDecimal", null, locale);`

PR #1404, opened 2026-07-02
create junit and integration test for rest api component (OFBIZ-13443)

Lukas-Finster:
Improved:  create junit and integration test for rest api component (OFBIZ-13443)
(OFBIZ-13443)

### DIFF
--- a/framework/rest-api/src/main/java/org/apache/ofbiz/ws/rs/model/ModelApiReader.java
+++ b/framework/rest-api/src/main/java/org/apache/ofbiz/ws/rs/model/ModelApiReader.java
@@ -25,6 +25,7 @@ import java.io.IOException;
 import javax.xml.parsers.ParserConfigurationException;
 
 import org.apache.ofbiz.base.util.Debug;
+import org.apache.ofbiz.base.util.UtilValidate;
 import org.apache.ofbiz.base.util.UtilXml;
 import org.w3c.dom.Element;
 import org.xml.sax.SAXException;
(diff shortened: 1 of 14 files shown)


Lukas-Finster:
Added JUnit tests for rest-api component and integration test for the auth service. Fixed an unhandled npe that got revealed while writing these, just highlighting the importance of unit tests in general.
Of note:
ApiRootResource#joinUri does not normalize double '/' in the current implementation. I am not 100% sure this is intended, some second thoughts would be appreciated. In the meantime implemented a test checking for this behaviour, to ensure a change to the normalization logic is a conscious one.

framework/rest-api/src/test/java/org/apache/ofbiz/ws/rs/core/ResponseStatusTest.java
```diff
@@ -0,0 +1,70 @@
...
+        assertSame(Custom.UNPROCESSABLE_ENTITY, result);
+    }
+
+    @Test
+    void testFromStatusCodeReturnsNullForUnknownStatusCode() {
+        assertNull(Custom.fromStatusCode(999));
+    }
+
+    @Test
+    void testFromStatusCodeReturnsNullForStandardStatuspCode() {
```
dixitdeepak: Minor typo in the method name.

framework/rest-api/src/main/java/org/apache/ofbiz/ws/rs/model/ModelApiReader.java
```diff
@@ -105,14 +106,19 @@ private static void createOperations(Element resourceEle, ModelResource resource
         for (Element operationEle : UtilXml.childElementList(resourceEle, "operation")) {
             Element serviceEle = UtilXml.firstChildElement(operationEle, "service");
             String serviceName = UtilXml.checkEmpty(serviceEle.getAttribute("name")).intern();
```
dixitdeepak: This can throw a NullPointerException if serviceEle is missing.

Since this change is intended to guard against an NPE in ModelApiReader when the `<service> ` element is missing from a ModelResource, we should also check that serviceEle is not null before accessing its attributes.

PR #1479, opened 2026-07-21
Improved: previous request information lost at login

nmalin:
Currently, when you arrive on OFBiz after the login page, you're redirected to the default request even though you have specified a URI.

Try to log on to https://localhost:8443/webtools/control/entity/find/Party, you are redirected to the login page, and after login, OFBiz returns to the page https://localhost:8443/webtools/control/main instead of the asked URI: entity/find/Party

This comes from issue OFBIZ-12047 (d60bf1574b, 33fe2aaba5) where we remove the previous request on the 'login' URI to be sure to clean it.

We reversed the logic, and set the previous request only when checkLogin controls a request that needs authentication and isn't an AJAX call or login (or related) URL.

### DIFF
--- a/framework/base/src/main/java/org/apache/ofbiz/base/util/UtilHttp.java
+++ b/framework/base/src/main/java/org/apache/ofbiz/base/util/UtilHttp.java
@@ -413,12 +413,14 @@ public final class UtilHttp {
 
     /**
      * Return pathInfo without parameters passed as query
-     * @param pathInfo
+     * @param request
      */
-    public static String getPathInfoWithoutQuery(String pathInfo) {
-        return pathInfo.indexOf('?') > -1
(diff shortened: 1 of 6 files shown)


jacopoc:
Hi @nmalin 
The issue you have identified may be a side effect of one of the fixes to security vulnerabilities I pushed this year.
However, this pull request (re?)introduces a security vulnerability (see the CodeQL report), so it cannot be pushed in its current form. Please have a look when you have time.

nmalin:
Hello @jacopoc,  Yes, I agree.

I try to find the good way on this spaghetti mix. I exposed my advancement on it and like you spot it, it's not enough.

I would like to secure the _PREVIOUS_REQUEST_ attribute at the beginning for that, but as it's a sensible area, I clearly wait for your (or some other) approval.

I needed to move forward on other things, but sure, I'll look into it to solve that.

PR #1511, opened 2026-07-29
Fixed: Time zones always displayed using their daylight savings display name (OFBIZ-12731)

toaditi:
[OFBIZ-12731](https://issues.apache.org/jira/browse/OFBIZ-12731)

This builds directly on work already done on the ticket. Credit where it is due:

- **Daniel Watford** reported the issue, suggested the `ZoneId#getDisplayName` approach back in 2022, and — after the footer was fixed — tested trunk and pinpointed that the `ListTimezones` page was the remaining part ("only partially fixed and shouldn't be closed").
- **Eugen Stan** authored the footer fix for OFBIZ-12721 (ca0433d0e4), which this change follows exactly.
- **Jacques Le Roux** followed up on both issues over the years and noticed that the time zone name shown in the footer could not be found in the list.

### The remaining problem

The time zone lookup list (`partymgr/control/ListTimezones`) still built its display names with `TimeZone#getDisplayName(useDaylightTime(), LONG, locale)`. `useDaylightTime()` answers whether the zone *has* DST rules at all, not whether DST is currently in effect, so every DST-observing zone was permanently shown with its summer name — e.g. "British Summer Time" in December, or "Australian Eastern Daylight Time" during the southern winter.

### The change

Use the same `ZoneId#getDisplayName(TextStyle.FULL_STANDALONE, locale)` call the footer has used since Eugen's OFBIZ-12721 fix, so the list shows the season-neutral generic name ("British Time", "Central European Time"). This also makes the list entries match the footer name, addressing Jacques's observation ("heure d'Europe centrale" is now findable in the list).

### Verification

On a local instance the rendered page shows 631 zones with no "Summer Time"/"Daylight Time" variants remaining; Europe/London → "British Time" (what Daniel saw in the fixed footer), Australia/Sydney → "Eastern Australia Time" (the previous code shows "Australian Eastern Daylight Time" today, mid-southern-winter), Europe/Paris → "Central European Time".

Supersedes #1510 (same change; reopened to properly credit the prior work in the commit message).

### DIFF
--- a/themes/common-theme/template/UserProfile.ftl
+++ b/themes/common-theme/template/UserProfile.ftl
@@ -27,7 +27,6 @@ under the License.
 <#if !(selectedTimeZoneId?has_content) && timeZone??>
   <#assign selectedTimeZoneId = timeZone.getID()>
 </#if>
-<#assign displayStyle = Static["java.util.TimeZone"].LONG>
 <#assign availableLocales = Static["org.apache.ofbiz.base.util.UtilMisc"].availableLocales()>
 <#assign availableTimeZones = Static["org.apache.ofbiz.base.util.UtilDateTime"].availableTimeZones()>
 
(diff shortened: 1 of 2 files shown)


dixitdeepak:
@toaditi 
Changes looks good to me, 
Could you please apply same fix in  UserProfile.ftl as well ?

toaditi:
Thank you for the review, Deepak! I've applied the same fix to UserProfile.ftl in da49e28 — both the current time zone row and the available time zones drop-down now use the generic ZoneId display name, and the unused displayStyle assign is removed. I verified on a local instance: with the time zone set to Australia/Sydney, the profile now shows "Eastern Australia Time (Australia/Sydney)" instead of the daylight name, matching the footer and the ListTimezones page. As far as I can tell this was the last remaining use of the useDaylightTime() display pattern in the framework.

PR #1606, opened 2026-08-09
Improved: Introduce a Gradle version catalog for dependency versions (OFBIZ-13487)

mridulpathak:
Extracts dependencies.gradle's ~98 scattered dependency-version literals into a new
gradle/libs.versions.toml Gradle version catalog, with zero change in resolved
dependency versions -- verified via an empty before/after dependency-tree diff and a
full build + test run. Plugin-local dependencies (e.g. plugins/ldap) are untouched;
the catalog covers only core-framework/shared dependencies. settings.gradle is also
untouched: its two plugin versions can't use the catalog, since Gradle evaluates that
file's plugins {} block before any version catalog exists (gradle/gradle#24876,
closed not-planned).

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -17,120 +17,109 @@
  * under the License.
  */
 dependencies {
-    implementation 'com.drewnoakes:metadata-extractor:2.20.0'
-    implementation 'com.github.ben-manes.caffeine:caffeine:3.2.4'
-    implementation 'com.google.guava:guava:33.6.0-jre'
-    implementation 'com.google.zxing:core:3.5.4'
-    implementation 'com.googlecode.concurrentlinkedhashmap:concurrentlinkedhashmap-lru:1.4.2'
-    implementation 'com.googlecode.ez-vcard:ez-vcard:0.12.2'
-    implementation 'com.googlecode.owasp-java-html-sanitizer:owasp-java-html-sanitizer:20240325.1'
-    implementation 'com.googlecode.libphonenumber:libphonenumber:8.13.55'
-    implementation 'com.ibm.icu:icu4j:76.1'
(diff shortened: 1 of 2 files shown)


jacopoc:
Hi @mridulpathak 

Did you check if, with the new format, Dependabot is still producing pull requests to keep the dependencies updated or if any configuration changes are required to our GitHub actions to allow it?
For example, I have noticed that Dependabot has recently created a few pull requests to bump Swagger in the rest-api component of the release branch (in which the component is under plugins, but this is irrelevant): see for example https://github.com/apache/ofbiz-plugins/pull/355
However I don't see an equivalent pull request for the framework trunk.

mridulpathak:
> Hi @mridulpathak
> 
> Did you check if, with the new format, Dependabot is still producing pull requests to keep the dependencies updated or if any configuration changes are required to our GitHub actions to allow it? For example, I have noticed that Dependabot has recently created a few pull requests to bump Swagger in the rest-api component of the release branch (in which the component is under plugins, but this is irrelevant): see for example [apache/ofbiz-plugins#355](https://github.com/apache/ofbiz-plugins/pull/355) However I don't see an equivalent pull request for the framework trunk.

Hi @jacopoc 

I ran Dependabot manually for build.gradle and it produced two PRS #1610 and #1611 with the new setup as expected. I noticed there are a lot of failed PR generations from before due to GitHub's default limit of 5. Would it be wise to configure a higher limit in depandabot.yml? Also, all the open framework dependabot PRs are now stale with new changes, I'll go ahead and close them.
LastRun
LastRunError

jacopoc:
Hi @mridulpathak 

I confirm that the Gradle catalog is monitored by Dependabot. This is good, please disregard my previous comment.
As regards the errors you have reported ("Dependabot cannot open any more pull requests"), I agree we should increase the limit. For now I am thinking to set it to 20 and see how it goes.
What do you think?

mridulpathak:
20 seems a reasonable limit to me, thanks @jacopoc.

PR #1609, opened 2026-08-09
Implemented: Do not expose services as REST endpoints when setting export=true in service definitions

Criux:
Implemented: Services whose definitions set export="true" no longer automatically also create a REST endpoint that exposes the service. The goal of this change is to:

* Reduce the risk that services are implicitly exposed as REST services, when this was not the original intent of the developer
* Unify the way REST endpoints are declared by creating a *rest.xml in an OFBiz component
* Make the implementation of the rest-api component simpler

Thanks to everyone that participated in the discussion about this change here: https://lists.apache.org/thread/jhqs0b3s3bkdd7rhgzpr131c3sfxsfxj


jacopoc:
Hi @Criux, the implementation looks good to me, thank you!

mridulpathak:
@Criux Nice cleanup. A few notes,

1. ApiServiceRequest and JsonifiedParamConverterProvider were only used by deleted OFBizServiceResource, so I think they could go too.
2. ServiceNotFoundException and MethodNotAllowedException were only used by deleted ServiceRequestFilter/ServiceRequestProcessor, they could go too.
3. OFBizResourceScanner.isIgnored() was ignoring only the deleted OFBizServiceResource, I think it can be removed including the static IGNORED Set.
4. OFBizOpenApiReaderTest.java:166 (`when(service.isExport()).thenReturn(true);`) can be deleted.

Criux:
Thank you everyone for reviewing the changes and especially thank you to @mridulpathak for your recommendations. I have added them to this PR as well.

PR #1644, opened 2026-08-13
Improve controller request/view resolution - suffix-selected

ashishvijaywargiya:
Tighten how suffix-selected views resolve against their owning request map, with related view-map adjustments and test coverage.


ashishvijaywargiya:
Thank you @Krishnauprit18 for you help.

PR #1649, opened 2026-08-14
Every ofbiz-launched dev server needs test classes on its classpath,

ashishvijaywargiya:
Every ofbiz-launched dev server needs test classes on its classpath, not just one started with --test/-t: webtools' "Run Test" screen (RunTestEvents -> TestRunContainer) loads jupiter-test-suite/junit-test-suite classes by name inside whatever JVM is already running, regardless of how that JVM was started. OFBIZ-13402 moved those classes out of src/main/* (and so out of sourceSets.main) to keep them out of release artifacts; this task is a local dev-run launcher, never a release-packaging task, so widening its classpath here doesn't reintroduce test code into anything that gets distributed.


ashishvijaywargiya:
Already committed the changes in a separate PR - https://github.com/apache/ofbiz-framework/pull/1655, so we can close this one.

PR #1694, opened 2026-08-21
Fixed: error in creating category image content. (OFBIZ-12011)

rehan3586:
Fixed: error in creating category image content
(OFBIZ-12011)

Explanation: the root cause was essentially a combination of incorrect parameter/field handling and missing ID propagation between the services involved in the category image upload process.

The fix ensures that the uploaded image is correctly received, its data resource is maintained, the generated content remains identifiable, and any service failure is properly surfaced

Thanks:
Aakash Jain for reporting the issue.

### DIFF
--- a/applications/product/src/main/groovy/org/apache/ofbiz/product/product/category/CategoryContentServicesScript.groovy
+++ b/applications/product/src/main/groovy/org/apache/ofbiz/product/product/category/CategoryContentServicesScript.groovy
@@ -24,6 +24,7 @@ import org.apache.ofbiz.base.util.UtilDateTime
 import org.apache.ofbiz.entity.GenericValue
 import org.apache.ofbiz.entity.util.EntityUtil
 import org.apache.ofbiz.service.ModelService
+import org.apache.ofbiz.service.ServiceUtil
 
 /**
  * Create Content For Product Category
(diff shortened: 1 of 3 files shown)


applications/product/src/main/groovy/org/apache/ofbiz/product/product/category/CategoryContentServicesScript.groovy
```diff
@@ -176,23 +177,33 @@ Map createDownloadContentForCategory() {
...
     contentMap.contentTypeId = 'DOCUMENT'
     Map creConRes = run service: 'createContentFromDataResource', with: contentMap
+    if (!ServiceUtil.isSuccess(creConRes)) {
+        return creConRes
+    }
     createCategoryContent.contentId = creConRes.contentId
 
-    createCategoryContent.contentId = parameters.contentId
-    run service: 'createCategoryContent', with: createCategoryContent
+    Map createCategoryResult = run service: 'createCategoryContent', with: createCategoryContent
```
vyasdevanshu, line 204: We should add service response handling here for the createCategoryContent call because if the service fails, the contentId will be added to the createCategoryResult, which is an error result.
Also, the contentId is optional=false in the service definition of this service, as it has to be returned properly in case of success.
Please check if this is properly handled here.

Abhishek-Kushwaha-1999: Hi @vyasdevanshu, Added check to handle the error if the service return.

rehan3586: @devanshu-vyas,

Thanks for the feedback. I have added the required condition to handle service error.

vyasdevanshu: Thanks.

PR #1701, opened 2026-08-23
Add batch test-run REST endpoint

ashishvijaywargiya:
Adds a batch endpoint on top of the existing test-run REST API in framework/testtools, fanning a
full-suite run out to multiple components in one call, tracked under one batchId:

  POST /rest/testtools/testruns/batch
  GET  /rest/testtools/testruns/batch/{batchId}

With no components list, every component that has a testdef and has the test execution API
enabled is queued automatically. With an explicit components list, an empty list is rejected
outright, and every named component must itself be known, have a testdef, and be enabled, or the
whole call is rejected up front naming which entries were invalid - no batch is ever queued from
an invalid or empty list. A component that passes this upfront check but still fails its actual
test run (e.g. a testdef that resolves to zero tests) is omitted from the batch rather than
failing the whole call, the same way an auto-discovered component would be. Each queued component
always runs its whole suite through the existing runTestSuite service, called as a plain Java
method, not through the service engine, so the batch stays serialized on the same single-threaded
executor runTestSuite already uses - no new concurrency anywhere in this change.

getBatchTestRunStatus polls a batch's aggregate status (QUEUED/RUNNING/PASSED/FAILED/ERROR,
rolled up from its children's own tracked statuses) plus each component's own runId/status/
resultSummary, reusing the same response shape GET testruns/{runId} already returns for that
runId.

No change to the existing single-component runTestSuite/getTestRunStatus contract, except one
additive field: getTestRunStatus's response now also includes runId alongside the existing
status/componentName/resultSummary.

Testing

- Unit tests for BatchRunTracker, BatchRunServices, and the extended TestRunServices: all
  passing, including status-rollup precedence (a batch with one ERROR and one FAILED child rolls
  up to ERROR, not FAILED, etc.), duplicate-component-list dedup, and rejecting a batch where
  every requested component's runTestSuite call fails outright rather than silently reporting
  PASSED with zero children.
- Full testIntegration run: all suites passing, zero regressions.
- Manual verification against a running server: auto-discovery batch across every enabled
  component, explicit-list batch, both 422 edge cases (an unknown component in an explicit list;
  the test execution API disabled globally), and the existing single-component endpoint's
  unaffected behavior plus the additive runId field.

### DIFF
--- a/framework/testtools/api/testruns.rest.xml
+++ b/framework/testtools/api/testruns.rest.xml
@@ -41,5 +41,14 @@ under the License.
         <operation verb="get" description="Get a test run's status" path="{runId}">
             <service name="getTestRunStatus"/>
         </operation>
+
+        <operation verb="post" description="Trigger a batch test-suite run across multiple components"
+                   path="batch" consumes="application/json">
+            <service name="runBatchTestSuite"/>
+        </operation>
+
+        <operation verb="get" description="Get a batch run's status" path="batch/{batchId}">
+            <service name="getBatchTestRunStatus"/>
+        </operation>
     </resource>
 </api>
--- a/framework/testtools/servicedef/services.xml
+++ b/framework/testtools/servicedef/services.xml
@@ -78,7 +78,7 @@ under the License.
             Exposed directly via the generic, framework-owned framework/testtools/api/testruns.rest.xml
             endpoint (GET /rest/testtools/testruns/{runId}) - same caution as runTestSuite above applies to
             any component-owned *.rest.xml.</description>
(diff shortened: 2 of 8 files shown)


jacopoc:
Hi @ashishvijaywargiya 

Could you please have a look at the following warning message generated by the code at line 330 of the TestRunServicesTest class? Thanks!

ashishvijaywargiya:
Thank you @jacopoc for letting me know about this. 

I just fixed the warning in the below PR:

https://github.com/apache/ofbiz-framework/pull/1801

PR #1717, opened 2026-08-24
Improved: authorization checks on Catalog and Order Manager helper routes

Krishnauprit18:
Improved: authorization checks on Catalog and Order Manager helper routes
- getChild and getProductStoreGroupRollupHierarchy in the Catalog Manager
  now require login and a CATALOG _VIEW permission check, matching the
  pattern used elsewhere in this package.
- Order Manager's quickadd request-map and view-map no longer carry a
  stray auth=false override, so it follows the webapp's normal login
  requirement like its sibling views.


ashishvijaywargiya:
@Krishnauprit18 thank you for your contribution on this issue.

PR #1736, opened 2026-08-26
Reject party/role combinations the party does not hold for BudgetRole (OFBIZ-12371)

karthikchundi-commits:
### Jira ticket
https://issues.apache.org/jira/browse/OFBIZ-12371

### Depends on #1727

**This PR depends on #1727 and should be reviewed/merged after it.** It builds directly on top of that branch, which adds the `checkPartyRoleExists` service this PR wires up for `createBudgetRole`. Without #1727, the `checkPartyRoleExists` service referenced here doesn't exist yet.

### Purpose and Description

Same bug shape as OFBIZ-12370/12372/12373 (fixed in #1727), one ticket over in the same family Pierre Smits filed: `createBudgetRole` has an unconditional `ensurePartyRole` `invoke` eca that silently fabricates a `PartyRole` record for whatever `roleTypeId`/`partyId` is submitted on the Budget Role page, with no validation that the party actually holds (or should hold) that role.

This wires the `checkPartyRoleExists` service added in #1727 as an `in-validate` eca on `createBudgetRole`, mirroring the exact same pattern already used there for `createInvoiceRole`/`createBillingAccountRole`/`createFinAccountRole`. Since `in-validate` runs before `invoke`, an invalid party/role combination is rejected before the existing `ensurePartyRole` eca ever fires - that eca is left untouched and becomes reachable only for combinations that are already valid.

### Changes 🏗️

- `applications/accounting/servicedef/secas.xml`: adds an `in-validate` eca on `createBudgetRole` calling `checkPartyRoleExists`, right before the existing `invoke` eca.
- `applications/accounting/src/test/groovy/org/apache/ofbiz/accounting/accounting/AutoAcctgBudgetTests.groovy`: adds a positive-case test (`DEMO_COMPANY` + `INTERNAL_ORGANIZATIO`, a role the test party already legitimately holds per `AccountingTestsData.xml`) and a negative-case test (`DEMO_COMPANY` + `CARRIER`, a role it does not hold - must now be rejected, and must not fabricate a spurious `PartyRole`), following the exact pattern used for the other three role types in #1727.

### Testing

- Verified `applications/accounting/servicedef/secas.xml` against the real `service-eca.xsd` schema via `xmllint --schema` - validates cleanly.
- Ran `gradle compileTestGroovy` (JDK 21) - compiles clean, including the new test methods.
- This repo's pre-push hook (`checkstyleMain checkstyleTest codenarcMain codenarcTest`) ran clean on push - no violations.
- I don't have a running OFBiz instance with a live database in this environment, so I could not execute the Groovy test suite itself end-to-end (only compile it) - same disclosed limitation as #1727.

### Checklist
- [x] My code follows the code style of this project
- [ ] I have updated the documentation accordingly (n/a - no user-facing docs for this internal validation)
- [x] I have read the CONTRIBUTING document
- [x] I have added tests to cover my changes

### DIFF
--- a/applications/accounting/servicedef/secas.xml
+++ b/applications/accounting/servicedef/secas.xml
@@ -118,6 +118,13 @@ under the License.
     </eca>
 
     <!-- budget role ecas -->
+    <!-- reject an improbable party/role combination before it reaches ensurePartyRole below,
+         which would otherwise silently create a spurious PartyRole - OFBIZ-12371 -->
+    <eca service="createBudgetRole" event="in-validate">
+        <condition field-name="roleTypeId" operator="is-not-empty"/>
+        <condition field-name="partyId" operator="is-not-empty"/>
+        <action service="checkPartyRoleExists" mode="sync"/>
+    </eca>
     <eca service="createBudgetRole" event="invoke">
         <action service="ensurePartyRole" mode="sync" run-as-user="system"/>
     </eca>
--- a/applications/accounting/servicedef/secas.xml
+++ b/applications/accounting/servicedef/secas.xml
@@ -47,6 +47,19 @@ under the License.
         <action service="updatePaymentMethodAddress" mode="sync"/>
     </eca>
 
+    <!-- reject an improbable party/role combination (party does not already hold roleTypeId) before it
+         reaches ensurePartyRole below, which would otherwise silently create a spurious PartyRole - OFBIZ-12370/12372/12373 -->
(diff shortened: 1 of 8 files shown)


karthikchundi-commits:
Following up on this one too — it depends on #1727 (which I just pinged), so flagging both together in case that's useful context for review order. Happy to make any changes needed. Thanks!

PR #1984, opened 2026-09-14
Fixed: PDF print and create file actions failing from Find Orders (OFBIZ-13505)

arunpati:
Fixes [OFBIZ-13505](https://issues.apache.org/jira/browse/OFBIZ-13505): selecting orders in Find Orders and running the Print (PDF) or Create File action produced no PDF, failing with `MissingPropertyException: No such property: session` and, once past that, FOP's "Document is empty" error.

## Root causes

The `OrderPDF` screen is rendered by the async services `sendPrintFromScreen` / `createFileFromScreen`, which have no web context:

1. `OrderViewWebSecure.groovy` unconditionally referenced `session` (permission checks) and `request` (content wrapper) — the reported `MissingPropertyException`.
2. `ScreenRenderer.populateContextForService` reads `userLogin` from the screenContext map, but callers only pass `orderId` there. With `userLogin` null, `hasPermission` stayed false, the permission-gated screen section rendered nothing, and FOP failed with "Document is empty".
3. The print path declared the job's DocFlavor as `application/postscript` while the FO transformation always produces PDF bytes — rejected by the OS print service (`PrintJobFlavorException: invalid flavor` on macOS).
4. The "Please select printer" placeholder option submitted the literal printer name `javascript:void(0);`.

## Changes

- **OrderViewWebSecure.groovy**: use the `userLogin` based `Security.hasEntityPermission` overload (the session overload just extracts userLogin, so web behavior is unchanged), and build the `OrderContentWrapper` with the dispatcher-based constructor when `context.request` is null — the same guard pattern `OrderView.groovy` already uses.
- **OutputServices.java**: pass `userLogin` from the service context into the screen render context in both `sendPrintFromScreen` and `createFileFromScreen`; default the print DocFlavor to `AUTOSENSE` instead of `application/postscript` (an explicitly passed `printerContentType` is still honored).
- **FindOrders.ftl**: empty value for the placeholder printer option, so the default printer is used instead of looking up a printer literally named `javascript:void(0);`.

## Testing

- Verified on a local instance with demo data: Find Orders → select orders → Create File now writes `order_<orderId>_<timestamp>.pdf` files with correct content; the normal backoffice order view, permission gating, and the ecommerce order view (IDOR guard, OFBIZ-11836) behave as before.
- `./gradlew classes` and the pre-push checks (checkstyle, codenarc) pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


arunpati:
Branch and PR should be created from fork, so closing this PR . Closed and deleted branch

PR #1994, opened 2026-09-14
Fixed: Preserve adjustments in quick order returns (OFBIZ-12432)

mohammad-k8:
Fixed:
(OFBIZ-12432)

Explanation:

quickReturnFromOrder did not pass orderAdjustmentId when copying order-level adjustments, so createReturnAdjustment produced zero-valued manual adjustments instead of preserving promotions, shipping charges, and taxes. The converted Groovy service also misspelled countNewReturnItems and read availableReturnTotal from the wrong service result, causing return reconciliation to use an incorrect value.

ReturnItems.ftl updated returnTotal only inside macro-local scope, so the return page displayed the raw item total without its item-level and return-level adjustments.

Pass the source adjustment IDs, use the correct service parameter and result, and accumulate adjustment amounts in the caller scope. Add a completed-order regression covering item and order promotions, shipping, tax, balance-adjustment suppression, and equal 17.00 return and invoice totals.

Thanks: @nameet-jain for reporting.


mridulpathak:
Thanks @mohammad-k8, fix looks right and I verified it locally (new test fails without the change, passes with it). Two requests:

1. Now that orderAdjustmentId is passed, createReturnAdjustment NPEs when the adjustment type has no ReturnItemTypeMap entry (DEPOSIT_ADJUSTMENT, DONATION_ADJUSTMENT, MKTG_PKG_AUTO_ADJUST), so a quick return on such an order fails outright. Could you add a null guard falling back to RET_MAN_ADJ?
https://github.com/apache/ofbiz-framework/blob/17736a72a4429359e9f86a0a5d7e056400a986bf/applications/order/src/main/java/org/apache/ofbiz/order/order/OrderReturnServices.java#L2445

2. Please keep the existing TEST_DEMO10090 testQuickReturnOrder and add the new assertions as a separate test, so that we don't loose the existing test coverage.

Same applies to #1995 (line 2454 on release24.09).

PR #2024, opened 2026-09-16
Fixed: PDF print and create file actions failing from Find Orders (OF…

arunpati:
Fixed: PDF print and create file actions failing from Find Orders (OFBIZ-13505)

Thanks for the report, Carsten Heinrigs

The OrderPDF screen is rendered by the async services sendPrintFromScreen and createFileFromScreen, which have no web context. Fixes:

- OrderViewWebSecure.groovy: use the userLogin based Security overload instead of the session one, and build the OrderContentWrapper with the dispatcher based constructor when no request is available. This removes the MissingPropertyException reported in the issue.
- OutputServices.java: pass userLogin into the screen render context so permission checks work, fixing the FOP "Document is empty" error, and default the print DocFlavor to AUTOSENSE since the FO transformation always produces PDF, not PostScript.
- FindOrders.ftl: use an empty value for the placeholder printer option so the default printer is used instead of a printer named "javascript:void(0);".


jacopoc:
@arunpati, could you please confirm whether this pull request, as well as pull request #2025, may have been merged before all CI/CD workflows had completed successfully?

In particular, it looks like the following workflow may not have completed successfully at the time the pull request was merged:
https://github.com/apache/ofbiz-framework/runs/104744323986

As a result, Scorecard has reopened the following alert:
https://github.com/apache/ofbiz-framework/security/code-scanning/1814

To help avoid similar issues in the future, could we please make sure that all CI/CD workflows have completed successfully before merging pull requests?

PR #2041, opened 2026-09-17
Fixed: Restore accounting entries invoice/payment PDF export (OFBIZ-12150)

mohammad-k8:
Fixed:
(OFBIZ-12150)

The Accounting Transactions Entries By Account screen retained an export link for a BIRT report after the BIRT component and its controller mappings were removed. As a result, the link opened a new window but could not resolve a view or render a PDF.

Restore the report with OFBiz's native screenfop renderer. Add the missing request and view mappings, reuse the accounting-entry search filters, and render the invoice, payment, account, party, currency, debit, and credit details without requiring BIRT.


mridulpathak:
Hi @mohammad-k8,

I tested it locally on trunk with demo data. Please address:

**Fixes**
- Exchange-rate column header renders as the raw key "FormFieldTitle_exchangeRate"; use the existing AccountingExchangeRate label.
- Exchange-rate column prints both amounts (e.g. "33.99/33.99 USD/USD") instead of a rate; compute origAmount / amount, and leave the cell blank when origCurrencyUomId equals currencyUomId.
- Add a debit and credit totals row, as in the original BIRT report.
- Update the description: BIRT wasn't removed, it lives in the birt plugin (disabled by default since OFBIZ-12824).

**Nice to have**
- Group rows by GL account class, as the original did.
- Show party names instead of party IDs.
- Widen the columns: they total ~316mm of ~406mm usable, leaving the right quarter of the page blank.

PR #2055, opened 2026-09-19
Fixed: EntityCrypto uses AES in ECB mode for encrypt="true" fields, producing deterministic ciphertext (OFBIZ-13599)

mridulpathak:
- Route EncryptMethod.TRUE through the same cipher default already used by SALT, instead of
  forcing AES into ECB mode (no IV, deterministic ciphertext for identical plaintext). Keep the
  old ECB cipher only as a decrypt fallback for values written before this fix.
- Retire encrypt="salt" as a separate option, now redundant with encrypt="true".
- Fix reencryptFields (OFBIZ-6694): select the primary key alongside the target field so store()
  can find the row afterward, and skip view entities, since storing through a view fails.
- Remove encrypt="true" from FinAccount.finAccountCode: the field is used for exact-match lookup
  as a redemption code (FinAccountHelper.getFinAccountFromCode), which is incompatible with
  non-deterministic encryption. finAccountPin remains encrypted. Add decryptFinAccountCodes, in
  the accounting component (not framework/entityext, to avoid a framework-to-application
  dependency), to migrate existing ciphertext values back to plain text.
- Remove the cardNumber search parameter from the findOrders service and admin order search,
  since exact-match search against an encrypted field is no longer possible.
- Document the required post-upgrade migration steps.


framework/entity/src/test/java/org/apache/ofbiz/entity/test/EntityCryptoTestSuite.java
```diff
@@ -145,19 +130,19 @@ public void testCryptoLookup() throws Exception {
...
-                "encryptedValue", nanoTime).queryList()).size());
+        // Regression guard for OFBIZ-13599: encrypted fields are no longer exact-match lookupable, since
+        // encryption is non-deterministic -- a freshly-encrypted search value never matches previously
+        // stored ciphertext for the same plaintext.
         assertEquals(0, (EntityQuery.use(delegator).from("TestingCrypto").where("testingCryptoTypeId", "LOOKUP",
-                "saltedEncryptedValue", nanoTime).queryList()).size());
+                "encryptedValue", nanoTime).queryList()).size());
 
-        assertEquals(1, EntityQuery.use(delegator).from("TestingCrypto").where("testingCryptoTypeId", "LOOKUP",
+        assertEquals(0, EntityQuery.use(delegator).from("TestingCrypto").where("testingCryptoTypeId", "LOOKUP",
```
jacopoc: The assertion at lines 145 and 146 is the same as the previous assertion (modified) at lines 142 and 143.

mridulpathak: Good catch, thanks — fixed in 9c131af5. That redundancy actually predated this PR (the original file had the same assertion twice among 5 lookup checks), it just became obvious once the encrypt="salt" ones were removed around it.

